PR #15, opened 2022-07-19
Bump packages package

mcasimir:
This is a proposal to move the `bump-package` script to its own package so we can automate Compass similarly to this repo.

- I would add the `.github/workflows/publish-packages.yaml` file from this repo in Compass as it is: packages with a version that is not present in the registry will be published from main.
- A pr will be created and updated with conventional bumps based on commits since the last publication.
- Unlike this repo, merging the PR will be manual, should not be necessary but is probably good to have a bit more control


packages/bump-monorepo-packages/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/mongodb-js/compass.git"
+  },
+  "files": [
+    "dist"
+  ],
+  "license": "SSPL",
+  "main": "dist/index.js",
+  "compass:main": "src/index.ts",
```
gribnoysup, line 29: These are webpack specific keys used in compass that we probably don't need here

packages/bump-monorepo-packages/src/bump-packages.js
```diff
@@ -1,14 +1,18 @@
+#!/usr/bin/env node
```
gribnoysup, line 1: I can't comment on the unchanged lines, but if we are planning to use this in compass, we will need to change this script

> // we increment the package version based on the bump on dependencies:
> // if a devDependency was bumped, regardless of the increment we increment of a
> // patch, otherwise we replicate the increment of the dependency. We always keep
> // the biggest increase.

Dependencies that are bundled with the package will be a dev dependency (it doesn't need to be explicitly installed when package is used, but it is part of distribution) and so assuming that the version bump can be a patch will be incorrect

gribnoysup (commented):
I am not sure what happens differently in this script compared to what lerna commands are already doing in compass monorepo. Maybe you can provide some context on this?

packages/bump-monorepo-packages/src/bump-packages.js
mcasimir: If I understand it correctly, if, for example, we change `compass-components` and add a new feature there, that would also be a new feature where is used, but `compass-components` is a development dependency of other packages and so the dependants would not be bumped as a feature bump, right?

I'm happy to find a solution for this, though I'm not too concerned, it would be rare that the work for a breaking change in a "library" package won't need changes on dependants too, that means you'd have the same "breaking" or "feat" commit also on the dependants which would bump them accordingly.

What would be a good alternative that doesn't need us to enter each version manually? Should we have the option to ignore the "type" of dependency? That sounds a bit drastic, for example, a breaking change in the eslint config would cause a major bump everywhere.

mcasimir: I wonder .. is there a way to detect those packages that are meant to be used as webpack? or worse case we can come up with something like a custom prop? what do you think?

gribnoysup: In compass we parse the source to detect what are the production dependencies of the package: https://github.com/mongodb-js/compass/blob/a43c2b20c67310ba845d88709a3e1b0339c421a2/scripts/check-peer-deps.js#L9-L42

I guess we can adapt the same thing here, this script does expect all the source to be in the `src` folder though, which we might need to adapt somehow, maybe use src by default but make it changeable through some package.json option or something

mcasimir:
> I am not sure what happens differently in this script compared to what lerna commands are already doing in compass monorepo. Maybe you can provide some context on this?

That's a good point, I think we discussed this already. Assuming we want to automate both bumping and publishing, the main difference is that this script doesn't use tags (also we can tweak the bump criteria but I didn't know we needed that till now), which let us do the package bump in a PR, review it and tweak it if we want.

In theory, we wouldn't need to announce publishing packages or anything (especially if and when we re-enable the requirement for the PRs to be up-to-date).

packages/bump-monorepo-packages/src/bump-packages.js
gribnoysup: Another thing here that I see is that this script is missing actual tagging and commiting part. Lerna doesn't support just running `version` to produce a commit with tags so if we are planning to use this script, we would need to move this logic here (or in an additional script) too

mcasimir: > Another thing here that I see is that this script is missing actual tagging and commiting part.

Yeah, that would be intentional, I'd literally do this: 
- https://github.com/mongodb-js/devtools-shared/blob/main/.github/workflows/bump-packages.yaml

And:

- https://github.com/mongodb-js/devtools-shared/blob/main/.github/workflows/publish-packages.yaml

No need to tag, unless we really want to, just to have a pointer to the commit where a bump happened, i don't think that's a requirement, we only did that cause lerna did that.

gribnoysup: I found tags super helpful to find and compare changes between releases, so ideally if it's not much effort I would try to keep them. This script seem to already have all the required info that you'd need to correctly tag a release and commit

gribnoysup:
Lerna can do verison bumps following conventional commits if you run `lerna version --conventional-commits`. It seems to even generate a pretty decent changelog by default (but we can disable this). At a glance results look pretty decent. If you're interested, you can try running

```
npx lerna version --conventional-commits --no-git-tag-version --no-push
```

in Compass monorepo to see the results. The CHANGELOGs are weird sometimes because since last release some packages had only `chore` commits, but that's the only thing that caught my eye. Even the CHANGELOG for packages/compass that I accidentally bumped looks very nice, although because we never actually changed the version there, lerna added like every change in the file 😆

mcasimir:
>  It seems to even generate a pretty decent changelog

Yeah, considering that we do that in a PR there shouldn't be any issue committing it .. buh .. if possible I'd rather disable that, seems something we don't need and the release notes we have right now look pretty good.

> ```
> npx lerna version --conventional-commits --no-git-tag-version --no-push
> ```

mmm ok, I think I'm onto something then, this would ignore the src thing though? cause the only issue I know with this command is that it only works using tags to find the base commit, but we can work that around but tagging the merge commit on main. Would we would have the issue with `src` and not bumping to major on breaking changes to devDeps?

packages/bump-monorepo-packages/src/bump-packages.js
gribnoysup: That would be fine, not critical, yeah. Just to clarify the flow I'm using sometimes is to just search for the tag in GitHub UI and then use the compare feature right there:

image

package.json
```diff
@@ -40,6 +40,7 @@
   "devDependencies": {
     "@babel/core": "7.16.0",
     "@babel/parser": "7.16.0",
-    "lerna": "^4.0.0"
+    "lerna": "^4.0.0",
+    "@mongodb-js/bump-monorepo-packages": "file:./packages/bump-monorepo-packages"
```
gribnoysup, line 44: Having it as a dev dep like that still seems a bit off to me, especially because in theory for workspaces it should be a version here and I'm worried that having a file path here might mess up the npm dependencies tree. In compass monorepo even without having this compass-scripts package that is part of the monorepo as a dependency in the root I am able to run them both with npx and with npm scripts:

image

image

gribnoysup (approved):
I would try to avoid having a file: dependency at the root, but other than that lgtm

package.json
mcasimir: interesting, i couldn't make it work, have to try again

PR #16, opened 2022-07-22
chore(scripts): cleanup scripts

mcasimir:
## Checklist


scripts/src/create-workspace.js
```diff
@@ -52,19 +56,18 @@ async function main(argv) {
...
     depType,
   } = await prompts(
     [
       {
         type: workspaceNameFromArgs ? null : 'text',
         name: 'name',
-        message: 'Provide a name for the new workspace',
-        hint: '(this name will be part of the package name)',
+        message: 'Provide a name for the new workspace (package name)',
+        hint: '(this name willProduction be part of the package name)',
```
gribnoysup, line 70:
```suggestion
        hint: '(this name will be part of the package name)',
```

scripts/src/create-workspace.js
```diff
@@ -103,37 +96,13 @@ async function main(argv) {
...
-            return null;
-          }
-
-          return 'confirm';
-        },
+        type: 'confirm',
         name: 'isPublic',
         message: 'Is it a public package?',
         initial: true,
       },
```
gribnoysup, line 103: Not something that we need to change right away, but similar to react, I was wondering if there ever will be a case for something to be a private package here? Even in Compass monorepo I don't think we used this option much

gribnoysup (approved):

scripts/src/create-workspace.js
mcasimir: Yeah, that makes sense .. why wouldn't we make a package public here, we already have `scripts` anyway

PR #34, opened 2022-10-12
chore(mongodb-constants): move shared constants from mongodb-ace-autoompleter package COMPASS-6173

gribnoysup:


packages/mongodb-constants/README.md
```diff
@@ -0,0 +1,21 @@
+# get-os-info
```
mabaasit, line 1: Maybe need to update readme?

gribnoysup: Whoops, thanks for catching!

gribnoysup: Fixed in https://github.com/mongodb-js/devtools-shared/pull/34/commits/a41288c83fc781551b27bc8bd257687a8cd12f07

packages/mongodb-constants/src/stage-operators.ts
```diff
@@ -0,0 +1,1014 @@
...
+    firstStage: false,
+    score: 1,
+    env: [ATLAS, ADL, ON_PREM],
+    meta: 'stage',
+    version: '3.4.0',
+    apiVersions: [1],
+    namespaces: [...ANY_NAMESPACE],
+    description:
+      'Automatically categorizes documents into a specified number of buckets, attempting even distribution if possible.',
+    comment: `/**
```
mabaasit, line 86: not sure if its important: once we have a bug in compass where stage comment was not valid and it required to update this package. maybe we can add test to validate if the comment is valid or not?

gribnoysup: Can you link to the case? This sounds like a reasonable thing to add, but I am not sure what valid comment would be / what to validate for here 🤔

mabaasit (approved):
looks good :)

packages/mongodb-constants/src/stage-operators.ts
gribnoysup: Aha, gotcha, thank you!

gribnoysup: Added a test https://github.com/mongodb-js/devtools-shared/pull/34/commits/f21f3d1676357f7497592c18f04462fd51064bfb

Does this make sense to you?

mabaasit: yes, that suffices!

PR #51, opened 2023-03-22
feat(mongodb-constants): add filter method

gribnoysup:
Extracting this from compass, as we want to use this in vscode too. One difference with compass implementation is that I removed the `prefix` filtering prop as this logic is not used by vscode and codemirror autocompleters, so it's only relevant for ace editor that we are planning to remove anyway

### DIFF
--- a/packages/mongodb-constants/src/filter.ts
+++ b/packages/mongodb-constants/src/filter.ts
@@ -1,0 +1,201 @@
+import { gte } from 'semver';
+import { ACCUMULATORS } from './accumulators';
+import { BSON_TYPE_ALIASES } from './bson-type-aliases';
+import { BSON_TYPES } from './bson-types';
+import { CONVERSION_OPERATORS } from './conversion-operators';
+import { EXPRESSION_OPERATORS } from './expression-operators';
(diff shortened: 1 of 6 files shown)


packages/mongodb-constants/src/filter.ts
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Convenience method to filter list of mongodb constants based on constants
+ * values
+ *
+ * @param options filter options
+ * @param constants list of constants to filter, for testing purposes only
+ * @returns filtered constants
+ */
+export function filter(
```
alenakhineika, line 168: I actually like the previous `completer` name more, because it not only filters methods, it returns the completion values. Maybe it can be `getFilteredComplitions` or something to point that it has a filter under the hood but also stores values?

gribnoysup: Sure, renamed. I think having a prefix match was making it more of a `completer`, now that this is removed, it's not really like that, so I went with `getFilteredCompletions` as it's somewhere in the middle

alenakhineika (approved):
Tried and love it :) 🚀

PR #67, opened 2023-05-18
fix: use nist.gov to collect CVEs scores for node

mcasimir:


packages/sbom-tools/src/commands/scan-node-js.ts
```diff
@@ -100,6 +72,58 @@ async function formatVuln(
...
+    const knownCvss: number[] = [];
+
+    for (const cvss of allCvss) {
+      if (typeof cvss === 'number') {
+        knownCvss.push(cvss);
+      }
+    }
+
+    return knownCvss.length ? Math.max(...knownCvss) : undefined;
+  } catch (e) {
```
lerouxb, line 119: The tiniest nit: I'd only try/catch around the `await Promise.all()` because that's where network /3rd party errors can happen. Now you're also doing try/catch around all our own potential runtime programming errors like "null pointer exceptions" (ie. reading something of undefined or whatever).

Although in this case that's probably not the end of the world because it would continue with undefined which (it seems) would be treated as the highest severity, so probably OK.

lerouxb (approved):

packages/sbom-tools/src/commands/scan-node-js.ts
lerouxb: ie. I'd usually prefer runtime programming errors to blow up the application just like it would outside of this try/catch.

mcasimir: that's a very good point, i'll update it

PR #70, opened 2023-05-22
chore(sbom-tools): add tests, readme and clean up COMPASS-6851

mcasimir:
Implements COMPASS-6851 cause non 3rd party modules are now never reported


packages/sbom-tools/test/helpers.ts
```diff
@@ -0,0 +1,28 @@
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
+import { rimrafSync } from 'rimraf';
+
+export const importFixture = async (file: string) =>
+  (await import(`./fixtures/${file}`)).default;
+
+export const toCleanup: string[] = [];
```
addaleax, line 9: Just a note, but it’s probably not the ideal way of doing things to have a global list here. In the tests for the products, we typically generate temporary directories in a `before` or `beforeEach` hook and then do the cleanup in a corresponding `after` or `afterEach` hook.

Also, these functions here seem to be used mostly from async functions or functions that could be made async functions fairly easily – maybe it’s a good idea to use normal fs functions rather than the sync ones? I know it doesn’t make a real difference for tests, but it’s a good practice in general and doesn’t really add extra code.

packages/sbom-tools/test/helpers.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function setupTempDir(structure: Record<string, string>): string {
+  const tempDirPath = path.join(os.tmpdir(), `sbom-tools-${Date.now()}`);
+
+  fs.mkdirSync(tempDirPath, { recursive: true });
+  toCleanup.push(tempDirPath);
+
+  for (const [entryPath, content] of Object.entries(structure)) {
+    const absoluteEntryPath = path.join(tempDirPath, ...entryPath.split('/'));
```
addaleax, line 18:
```suggestion
    const absoluteEntryPath = path.join(tempDirPath, entryPath);
```

packages/sbom-tools/src/webpack-dependencies-plugin.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+  const contextPath = setupTempDir(structure);
+
+  const outputPath =
+    options.outputFilename ?? path.join(contextPath, 'dependencies.json');
+  toCleanup.push(outputPath);
+
+  return new Promise((resolve, reject) => {
+    const plugin = new WebpackDependenciesPlugin(options);
+    const config = fs
+      .readdirSync(path.join(contextPath, './src'))
```
addaleax, line 23:
```suggestion
      .readdirSync(path.join(contextPath, 'src'))
```

packages/sbom-tools/src/webpack-dependencies-plugin.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+        context: contextPath,
+        entry: path.resolve(contextPath, 'src', filename),
+        output: {
+          path: path.resolve(contextPath, 'dist'),
+          filename: `output-${i}.js`,
+        },
+        plugins: [plugin],
+      }));
+
+    webpack(config, (err, stats) => {
```
addaleax, line 37: This looks like something worth `util.promisify()`ing and then just using it with `await`

addaleax (approved):
(All comments completely optional!)

packages/sbom-tools/test/helpers.ts
mcasimir: We can improve this yeah, i thought it is fine in this context, those tests should not run in parallel, but yeah, we can code this properly

packages/sbom-tools/test/helpers.ts
mcasimir: Is that ok in win? I'm always adding entries like this: `{'node_modules/my-file.js': '....'`}

addaleax: Yes, `/` is also a path separator on Windows (basically aliases to `\`), and Node.js also does understand this properly:

```js
> path.win32.join('a', 'b/c/d/')
'a\\b\\c\\d\\'
> path.posix.join('a', 'b/c/d/')
'a/b/c/d/'
```

packages/sbom-tools/test/helpers.ts
addaleax: It is fine, and again, this is optional, it’s just not really matching what our tests usually do, and the way that our tests usually do it does have its advantages. But yeah, not a big deal unless we start copying this kind of code around

packages/sbom-tools/test/helpers.ts
mcasimir: awesome, thought that the separator was always different, thanks!

packages/sbom-tools/test/helpers.ts
mcasimir: yeah, no is a very good point, let's do it in a way that is up to our standard, is a quick thing to change it

mcasimir: Updated, let me know if that's ok

addaleax (approved):

PR #85, opened 2023-06-15
feat(oidc-mock-provider): add bin script for running mock identity provider server

Anemy:
Also adds an option `hostname` to the config options.


packages/oidc-mock-provider/bin/oidc-mock-provider.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+(async function main() {
+  const oidcMockProviderConfig = {
+    getTokenPayload() {
+      return DEFAULT_TOKEN_PAYLOAD;
+    },
+    overrideRequestHandler() {
+      return () => {};
+    },
```
addaleax, line 43:
```suggestion
```

(This is optional, and it doesn’t really matter that much either way, but it doesn’t return a function so I was going to leave a suggestion anyway :slightly_smiling_face:)

addaleax (approved):

PR #87, opened 2023-06-15
feat(oidc-mock-provider): add bindIpAll config option

Anemy:
Add `bindIpAll` config option that makes the server bind to both the host and `0.0.0.0` to hopefully workaround docker loopback and https requirement unless localhost for OIDC. This might not work, but saves some more complicated self signed certificate https configuration.

### DIFF
--- a/packages/oidc-mock-provider/src/index.ts
+++ b/packages/oidc-mock-provider/src/index.ts
@@ -54,6 +54,13 @@ export interface OIDCMockProviderConfig {
    * Optional hostname for the server to listen on.
    */
   hostname?: string;
+
+  /**
+   * For testing with docker instances it's useful
+   * to bind both to the local loopback address and to
+   * all of the ips so that we can avoid the https requirements.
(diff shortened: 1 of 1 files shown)


packages/oidc-mock-provider/src/index.ts
```diff
@@ -87,6 +94,9 @@ export class OIDCMockProvider {
 
   private async init(): Promise<this> {
     this.httpServer.listen(this.config.port ?? 0, this.config.hostname);
+    if (this.config.bindIpAll) {
+      this.httpServer.listen(this.config.port ?? 0, '0.0.0.0');
```
addaleax, line 98: Should the two listen calls be mutually exclusive?

> The `server.listen()` method can be called again if and only if there was an error during the first `server.listen()` call or `server.close()` has been called

Anemy: Oh...

PR #88, opened 2023-06-20
chore: factor out shared eslint config, publish to npm MONGOSH-1481

lerouxb:


configs/eslint-config-devtools/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.9.3",
   "description": "Shared DevTools eslint configuration",
   "license": "SSPL",
-  "private": true,
```
lerouxb, line 6: Hoping this is all I need so this will be published to npm

configs/eslint-config-devtools/common.js
```diff
@@ -0,0 +1,130 @@
+const tsRules = {
```
addaleax, line 1:
```suggestion
'use strict';
const tsRules = {
```

configs/eslint-config-devtools/index.js
```diff
@@ -1,67 +1,4 @@
...
-    ecmaFeatures: {
-      jsx: true,
-    },
-  },
-};
-
-// The one that the library comes with doesn't allow for numbers and files
-// starting with dots, this modified version handles those
-const kebabcase = /^\.?([a-z0-9]+-)*[a-z0-9]+(?:\..*)?$/;
+const common = require('./common');
```
addaleax, line 1:
```suggestion
'use strict';
const common = require('./common');
```

addaleax (approved):

PR #102, opened 2023-06-23
feat: add mongodb-runner, mongodb-downloader MONGOSH-1487

addaleax:
## Checklist


packages/mongodb-runner/src/mongocluster.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  after(async function () {
+    await fs.rm(tmpDir, {
+      recursive: true,
+      maxRetries: 100,
+    });
+  });
+
+  afterEach(async function () {
+    await cluster?.close();
```
lerouxb, line 32: Nitpick, but if a test fails at cluster = await MongoCluster.start()  then won't afterEach will re-close the previous cluster? Kinda inconsequential because a test would have already failed.

lerouxb: Actually nevermind. I see close() clears out all the servers anyway.

lerouxb (approved):

PR #103, opened 2023-06-26
feat: add @mongodb-js/monorepo-tools

lerouxb:
@mcasimir pointed out that publishing scripts/* isn't a good idea because that's for this repo's own scripts.

I also realised that the code that calculates MONOREPO_ROOT and LERNA_BIN breaks when you try and run the scripts from compass, so that had to be fixed.

We have many copies of the same utils and scripts, so this consolidates it all in a @mongodb-js/monorepo-tools package. That's useful not just for the scripts, but it also makes the functions reusable.

Once this is merged we can update compass (and others?) to use this. Can also move the scripts/monorepo in compass here. And remove duplicate scripts and utils in compass.


packages/monorepo-tools/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "type": "git",
+    "url": "https://github.com/mongodb-js/devtools-shared.git"
+  },
+  "license": "SSPL",
+  "files": [
+    "src",
+    "!src/**/*.spec.js",
+    "!src/fixtures"
+  ],
+  "main": "dist/index.js",
```
lerouxb, line 24: Copy/pasted this from bump-monorepo-packages, but now not sure this is right. What's making dist/?

packages/monorepo-tools/where.js
```diff
@@ -52,7 +53,7 @@ async function lernaExec(packages) {
     return;
   }
 
-  const lernaBin = path.resolve(process.cwd(), 'node_modules', '.bin', 'lerna');
+  const lernaBin = path.resolve(MONOREPO_ROOT, 'node_modules', '.bin', 'lerna');
```
lerouxb, line 56: I need to test the useLernaExec=true path still..

scripts/package.json
```diff
@@ -5,21 +5,18 @@
     "name": "MongoDB Inc",
     "email": "compass@mongodb.com"
   },
+  "private": true,
```
lerouxb, line 8: Taking this private again. Will unpublish once this is merged.

packages/monorepo-tools/utils/constants.js
```diff
@@ -0,0 +1,10 @@
+const path = require('path');
+const findUp = require('find-up');
+
+exports.MONOREPO_ROOT = path.dirname(
```
mcasimir, line 4: When used outside of this repo this will point to whatever package.json is found starting from 2 folders above the `node_modules` folder this file is, i'm not sure this is what we want.

If we want't to find the `monorepo` root from cwd, the safest way i can think of is to  `findUp.sync('package-lock.json', {cwd: process.cwd()})` and double-check that it contains workspace info.

mcasimir (changes requested):

packages/monorepo-tools/where.js
mcasimir: Why is this still needed?

mcasimir: we get packages as argument already:

``` js
for packageInfo of packages {
   await util.promisify(execFile)(execCommandArgs[0], ...execCommandArgs.slice(1), {cwd: packageInfo.location})
}
```

packages/bump-monorepo-packages/src/bump-packages.js
```diff
@@ -5,7 +5,7 @@ const path = require('path');
 const childProcess = require('child_process');
 const assert = require('assert');
 const { promisify } = require('util');
-const getPackagesInTopologicalOrder = require('./get-packages-in-topological-order');
+const { getPackagesInTopologicalOrder } = require('@mongodb-js/monorepo-tools');
```
mcasimir, line 8: NIT: do we want to move this package code inside the other one? this seems to be also a monorepo tool type of package

lerouxb: That's what I asked in DMs and in the PR description.

packages/monorepo-tools/utils/constants.js
lerouxb: We use MONOREPO_ROOT in 6 places so I thought the easiest way to get past review would be to try and make a drop-in replacement. I don't mind changing them all.

packages/monorepo-tools/where.js
lerouxb: I'm not sure what you're referring to. I just ported over the script and it has a path where it does what you copy/pasted and one where it uses lerna.

packages/monorepo-tools/utils/constants.js
lerouxb: I removed this file now.

packages/monorepo-tools/utils/find-monorepo-root.js
```diff
@@ -0,0 +1,14 @@
+const findUp = require('find-up');
+const fs = require('fs');
+const path = require('path');
+
+async function findMonorepoRoot() {
+  const packagePath = await findUp('package.json');
```
mcasimir, line 6: we can't find any `package.json` it would fail always when called from a script in a package.json of  a workspace package, which is a use case for the scripts.

find-up should let us pass a fn to check if the we like its content, or you can look for `package-lock.json` instead, basically saying  that is safer to assume it is only in the monorepo root

packages/monorepo-tools/package.json
lerouxb: I'll just convert all this type typescript.

packages/monorepo-tools/package.json
```diff
@@ -1,64 +1,67 @@
...
     "!src/**/*.spec.js",
     "!src/fixtures"
   ],
-  "license": "Apache-2.0",
-  "main": "dist/index.js",
+  "main": "index.js",
+  "bin": {
+    "precommit": "./precommit.js",
+    "depalign": "./depalign.js",
+    "monorepo-where": "./where.js",
```
lerouxb, line 28: I named the script monorepo-where because `where` is a builtin bash command.

mcasimir (approved):

PR #112, opened 2023-07-04
feat(mongodb-runner): add CLI script MONGOSH-1488

addaleax:
#102 + one commit

### DIFF
--- a/packages/mongodb-runner/README.md
+++ b/packages/mongodb-runner/README.md
@@ -5,7 +5,7 @@ Helper for spinning up MongoDB servers and clusters for testing.
 ## Example usage
 
 > Note: Version 5 of mongodb-runner is a full re-write. Many things work
-> differently in previous versions.
+> differently in version 5 and above.
 
 ```bash
 $ npx mongodb-runner start -t sharded
--- a/packages/mongodb-runner/README.md
+++ b/packages/mongodb-runner/README.md
@@ -1,0 +1,68 @@
+# mongodb-runner
+
+Helper for spinning up MongoDB servers and clusters for testing.
+
+## Example usage
+
+> Note: Version 5 of mongodb-runner is a full re-write. Many things work
+> differently in previous versions.
+
+```bash
+$ npx mongodb-runner start -t sharded
+$ npx mongodb-runner start -t replset -- --port 27017
+$ npx mongodb-runner start -t replset -- --setParameter allowDiskUseByDefault=true
+$ npx mongodb-runner stop --all
+$ npx mongodb-runner exec -t standalone -- sh -c 'mongosh $MONGODB_URI'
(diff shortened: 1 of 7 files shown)


packages/mongodb-runner/README.md
```diff
@@ -0,0 +1,68 @@
+# mongodb-runner
+
+Helper for spinning up MongoDB servers and clusters for testing.
+
+## Example usage
+
+> Note: Version 5 of mongodb-runner is a full re-write. Many things work
+> differently in previous versions.
```
lerouxb, line 8: Total copywriting nitpick, totally just ignore me: Feels like normally you'd word this sentence that it works different _now_ not in the past.

packages/mongodb-runner/README.md
```diff
@@ -0,0 +1,68 @@
...
+  mongodb-runner start  Start a MongoDB instance
+  mongodb-runner stop   Stop a MongoDB instance
+  mongodb-runner ls     List currently running MongoDB instances
+  mongodb-runner exec   Run a process with a MongoDB instance (as MONGODB_URI
+                        env var)
+
+Options:
+  -t, --topology     Select a topology type
+           [choices: "standalone", "replset", "sharded"] [default: "standalone"]
+      --arbiters     number of arbiter nodes for each replica set       [number]
```
lerouxb, line 34: Minor thing for all these numbers: Should we specify the default? Is there even a default in every case? I suppose each one only matters for specific code paths.

packages/mongodb-runner/src/cli.ts
```diff
@@ -1,31 +1,203 @@
...
+      ...argv,
+      args: mongodArgs,
+    });
+    try {
+      const [prog, ...progArgs] = execArgs;
+      const child = spawn(prog, progArgs, {
+        stdio: 'inherit',
+        env: {
+          ...process.env,
+          // both spellings since otherwise I'd end up misspelling these half of the time
```
lerouxb, line 181: 😄

lerouxb (approved):

packages/mongodb-runner/README.md
addaleax: So, just for context, the `default:` in yargs sets the actual default, it’s not just informative for the user … so setting e.g. `default: 2` for `secondaries` means that we’d maintain the default in two places. That might be a bit nicer for UX, but I’m not sure if it’s really worth it?

packages/mongodb-runner/README.md
addaleax: Sure, rephrased this a bit :+1:

packages/mongodb-runner/README.md
lerouxb: Yeah we're the only users, so probably not worth it.

PR #126, opened 2023-07-07
chore: run prettier via a script so we only have to specify the dependency once per monorepo MONGOSH-1482

lerouxb:
This came out of a comment by @gribnoysup on the mongosh prettier PR.

The idea is that we won't have to specify the prettier dep everywhere we use @mongodb-js/prettier-config-devtools. Ideally we'd specify the dependency only once there. This isn't exactly that, but at least now we only have to specify it once per monorepo. The nice thing is that it is still a peer dep, so should in theory be enforced? Also the precommit script and this script both does `npx prettier` so should pick up the same version, unlike before.

Hopefully this works in compass, mongosh and oidc-plugin. Can't fully test it until we publish, unfortunately.


configs/prettier-config-devtools/bin/monorepo-prettier
```diff
@@ -0,0 +1,13 @@
...
+# everywhere we specify @mongodb-js/prettier-config/devtools as a dep
+
+# Why npx? The current directory is probably in a different package, probably a
+# different monorepo rather than this one. So that's where we're executing
+# prettier from. And we can't mess with the current directory, because the
+# command-line args assume we're there. But we don't want to have to install
+# prettier everywhere so the best we can do is probably to specify prettier once
+# per monorepo at the top.
+
+npx prettier "$@"
```
gribnoysup, line 13: I think this is already pretty good, but if you want to avoid having prettier even at the monorepo root and make sure that you use the one that will be installed with this package, you can do something similar to [what we do in webpack-config-compass](https://github.com/mongodb-js/compass/blob/3813e8433f30460bc7d36ad98a6596be4d825197/configs/webpack-config-compass/bin/webpack.js#L36)

gribnoysup (approved):
Even if we don't manage to do the same for eslint, I think it's worth doing this for prettier, good stuff!

PR #128, opened 2023-07-10
bump @typescript-eslint to match compass

lerouxb:
compass is ahead of devtools-shared here and eslint being weird plus mixed versions and hoisting (or the lack thereof) is complicating matters for me on mongosh too. Easiest to just move this forward.


configs/eslint-config-devtools/common.js
```diff
@@ -16,6 +16,7 @@ const tsRules = {
   '@typescript-eslint/no-unsafe-call': 'off',
   '@typescript-eslint/no-unsafe-member-access': 'off',
   '@typescript-eslint/no-unsafe-return': 'off',
+  '@typescript-eslint/no-unsafe-argument': 'warn',
```
lerouxb, line 19: This became an error by default and there are a LOT of cases of any being passed around.

mcasimir (approved):

PR #145, opened 2023-07-28
feat(saslprep): add package COMPASS-6996

addaleax:
This is the current source of https://github.com/reklatsmasters/saslprep, with only the necessary modifications to align this package with the other devtools-shared repository (which does include moving it to TypeScript and switching from jest to mocha/chai).

## Checklist


packages/saslprep/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "url": "https://github.com/mongodb-js/devtools-shared.git"
+  },
+  "files": [
+    "dist"
+  ],
+  "license": "MIT",
+  "exports": {
+    "require": "./dist/index.js",
+    "import": "./dist/.esm-wrapper.mjs"
+  },
```
nbbeeken, line 33: If you use ESM, then you don't get type definitions. I think we need to make this similar to BSON: https://github.com/mongodb/js-bson/blob/main/package.json#L80-L91

packages/saslprep/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+    throw new Error(
+      'Bidirectional RandALCat character must be the first and the last' +
+        ' character of the string, see https://tools.ietf.org/html/rfc3454#section-6'
+    );
+  }
+
+  return normalized_input;
+}
+
+export = saslprep;
```
nbbeeken, line 152: Could this be a named or default export now? or is that too much friction?

packages/saslprep/package.json
addaleax: Fwiw, this is something that looks the same in every package.json in our repos here, we haven’t been seeing any issues ourselves :slightly_smiling_face:

addaleax: (But still happy to address this, of course, done in c261b1e9e7c6ddfd6b2f838ff2b54ad61039ee74)

packages/saslprep/src/index.ts
addaleax: We could do that, or we could add something like

```suggestion
saslprep.saslprep = saslprep;
saslprep.default = saslprep;
export = saslprep;
```

for compatibility. No strong feelings here, I mostly just wanted to keep everything as-is in terms of code and behavior for this PR.

packages/saslprep/package.json
nbbeeken: I tried it locally with the module: node16 setting, I think it's specifically that setting that causes the issue 🤷🏻 ty!

packages/saslprep/src/index.ts
nbbeeken: Fine with either approach, I know for us at least migrating to a named export will be easy even if we want to backport the migration to this dependency to previous driver versions. Changing this makes it easier to not use tslib/helpers which we are still avoiding (I can't speak for the future)

packages/saslprep/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "import": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/.esm-wrapper.mjs"
+    },
+    "require": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
+  },
+  "types": "./dist/index.d.ts",
```
baileympearson, line 40: question: what's the difference between `exports.import.types` and the `types` field?  Is there a reason we need both?

addaleax: See @nbbeeken’s first comment thread here: https://github.com/mongodb-js/devtools-shared/pull/145#discussion_r1277590030

Happy to revert the change again, it doesn’t make a difference for us :)

packages/saslprep/src/index.ts
addaleax: Ok, then I’ve made this change here as suggested above, I think that should be fine as an addition to the initial feature :+1:

nbbeeken (approved):

alenakhineika (approved):

PR #148, opened 2023-08-21
feat: add monorepo-tools bin as lerna replacement

mcasimir:
@lerouxb not sure if this is too much, the filter part in getPackagesByTopology is probably ok, since we are maintaining it already the added code + tests doesn't seem to be overwhelming, but do we want to have an actual lerna replacement? or something like  `monorepo-packages | xargs xargs -I{} npm run -w {}` (if it works everywhere)

### DIFF
--- a/packages/monorepo-tools/bin/monorepo-tools.js
+++ b/packages/monorepo-tools/bin/monorepo-tools.js
@@ -1,0 +1,3 @@
+#!/usr/bin/env node
+'use strict';
+require('../dist/monorepo-tools.js');
--- a/packages/monorepo-tools/src/monorepo-tools.ts
+++ b/packages/monorepo-tools/src/monorepo-tools.ts
@@ -1,0 +1,321 @@
+/* eslint-disable no-console */
+import { spawn } from 'child_process';
+import split2 from 'split2';
+
+import { findMonorepoRoot } from './utils/find-monorepo-root';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+import type { PackageInfo } from './utils/get-packages-in-topological-order';
+import { getPackagesInTopologicalOrder } from './utils/get-packages-in-topological-order';
+import chalk from 'chalk';
+
+type YargsCommand = ReturnType<typeof yargs>;
+type YargsOptionDefinition = Parameters<YargsCommand['options']>[0];
+type CommandOptions<T> = { _: string[] } & T;
+
+const getColorFn = (() => {
+  let i = 0;
+  const chalkColors = [
+    chalk.red,
+    chalk.green,
+    chalk.yellow,
+    chalk.blue,
+    chalk.magenta,
+    chalk.cyan,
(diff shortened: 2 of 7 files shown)


packages/monorepo-tools/bin/monorepo-tools.js
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
```
addaleax, line 1: This file is missing the +x flag, either the flag should be added or this line should be removed (I’d strongly lean towards the former :slightly_smiling_face:)

packages/monorepo-tools/src/monorepo-tools.ts
```diff
@@ -0,0 +1,321 @@
+/* eslint-disable no-console */
+import { spawn } from 'child_process';
+import split2 from 'split2';
```
addaleax, line 3: Up to you whether you want to keep this, but these days `split2` (for splitting on line endings) is something like a 8-line utility:

```js
function split2(map: (l: string) => string) {
  const input = new PassThrough();
  return Duplex.from({
    writable: input,
    readable: Readable.from((async function* () {
      for await (const line of /*readline.*/createInterface({ input }))
        yield map(line);
    })())
  });
}
```

packages/monorepo-tools/src/utils/get-packages-in-topological-order.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+      'config',
+      '--local',
+      'user.email',
+      'user@example.com',
+    ]);
+
+    // setup repo and package:
+    repoPath = path.resolve(tempDir, 'monorepo-test-repo');
+    await fs.mkdir(repoPath, { recursive: true });
+    process.chdir(repoPath);
```
addaleax, line 43: Should this be reset to its original value somewhere so it doesn’t mess with other tests (or even other test files) potentially added in the future?

(I’d personally try to avoid `chdir()` as much as possible, I don’t think having a `let cwd = remoteDir` at the start here and then doing `await execFile(..., { cwd });` each time would be that bad)

addaleax (approved):

packages/monorepo-tools/src/utils/get-packages-in-topological-order.spec.ts
mcasimir: sure! i've copy-pasted this from other compass tests and probably forgot few lines

mcasimir: fun how many bugs i've fixed by switching to { cwd }

addaleax (approved):

lerouxb (approved):

lerouxb:
I think you might as well replace (and remove) lerna inside the root package.json for all of devtools-shared in this PR. Would act as another "test" to prove that it works.

mcasimir:
Going to close this one

PR #153, opened 2023-08-24
chore: add a script to sort workspaces in package.json if workspacesPatterns exists

lerouxb:
In a monorepo that isn't also a library (ie. compass or mongosh) we can add a "postinstall" or "prepare" or similar npm install lifecycle hook that will run after `npm install` finishes that then makes sure that, if you have a workspacesPatterns property, the "workspaces" property is complete (ie. no globs) and in topological order.

Then we can do things like `npm run compile --workspaces` and npm will run compile in the correct order. Which is handy for moving off lerna in CI because newer versions of lerna don't support exotic platforms.

I didn't run this from a lifecycle hook in this monorepo yet because it gets a bit chicken and egg with the fact that we want to run it before compile but it is written in typescript and therefore has to be compiled first.. Solvable but probably messy?


package.json
```diff
@@ -12,11 +12,27 @@
     "node": ">=14.17.5",
     "npm": ">=7.16.0"
   },
-  "workspaces": [
+  "workspacesPatterns": [
     "packages/*",
     "configs/*",
     "scripts"
   ],
+  "workspaces": [
```
lerouxb, line 20: And then this script will manage the workspaces prop based on workspacesPatterns.

packages/monorepo-tools/src/utils/get-packages-in-topological-order.ts
```diff
@@ -17,20 +17,27 @@ export interface PackageInfo {
...
-      .workspaces || [];
+  if (!patterns) {
+    patterns =
+      JSON.parse(await fs.readFile(path.join(cwd, `package.json`), 'utf8'))
+        .workspaces || [];
+  }
 
   const packageJsonPaths = await glob(
-    patterns.map((pattern) => path.join(pattern, 'package.json'))
+    // NOTE: glob patterns should always use forward slashes,
```
lerouxb, line 30: Noticed this in @mcasimir 's PR and copied it here.

packages/monorepo-tools/src/sort-workspaces.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  if (!patterns) {
+    console.log('Refusing to sort workspaces without workspacesPatterns in monorepo root package.json');
+    return;
+  }
+
+  const packages = await getPackagesInTopologicalOrder(monorepoRoot, patterns);
+
+  packageJSON.workspaces = packages.map(({ location }) =>
+    location.replace(`${monorepoRoot}/`, '')
```
lerouxb, line 31: Actually thinking about this now, this will probably not work on windows 🤔

package.json
```diff
@@ -12,11 +12,27 @@
...
+    "packages/monorepo-tools/configs/prettier-config-devtools",
```
lerouxb, line 21: And I just noticed that my find/replace at the last minute is totally broken ;)

packages/monorepo-tools/src/sort-workspaces.ts
```diff
@@ -0,0 +1,45 @@
...
+});
+
+async function main() {
+  const monorepoRoot = await findMonorepoRoot();
+  const text = await fs.readFile(
+    path.join(monorepoRoot, 'package.json'),
+    'utf8'
+  );
+  const packageJSON = JSON.parse(text);
+  const patterns: string[] | undefined = packageJSON.workspacesPatterns
```
addaleax, line 21: The linter should complain here, right? :)

packages/monorepo-tools/src/sort-workspaces.ts
addaleax: Maybe https://nodejs.org/api/path.html#pathrelativefrom-to does the tr
ick here?

lerouxb: Turns out if I copy all the fixes it is already the path component returned by glob which is exactly what we need.

lerouxb:
Closing in favour of just using lerna to generate the list in mongosh since this is only useful if we're getting rid of lerna completely.

PR #154, opened 2023-09-04
chore: take monorepoRoot into account in getPackagesInTopologicalOrder()

lerouxb:
Found while working on https://github.com/mongodb-js/mongosh/pull/1621. `cwd` is potentially something quite deep and different from monorepoRoot when running `npm run evergreen-release bump` and then glob() ends up returning a blank array.

### DIFF
--- a/packages/monorepo-tools/src/utils/get-packages-in-topological-order.ts
+++ b/packages/monorepo-tools/src/utils/get-packages-in-topological-order.ts
@@ -11,9 +11,9 @@ interface InternalPackageInfo {
 
 export interface PackageInfo {
   name: string;
-  location: string;
   version: string;
   private: boolean;
+  location: string;
 }
 
 export async function getPackagesInTopologicalOrder(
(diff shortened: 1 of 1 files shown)


packages/monorepo-tools/src/utils/get-packages-in-topological-order.ts
```diff
@@ -17,24 +17,29 @@ export interface PackageInfo {
...
     packageJsonPaths.map(async (packageJsonPath) => {
+      const packageJsonLocation = path.resolve(monorepoRoot, packageJsonPath);
       const packageJson = JSON.parse(
-        await fs.readFile(packageJsonPath, 'utf8')
+        await fs.readFile(packageJsonLocation, 'utf8')
       );
       return {
         name: packageJson.name,
-        location: packageJsonPath,
+        location: path.dirname(packageJsonLocation),
```
addaleax, line 42: This was previously a filename and is now a directory name, just making sure that that is intentional?

addaleax (approved):

packages/monorepo-tools/src/utils/get-packages-in-topological-order.ts
lerouxb: Kinda. We only ever used it as a dirname and by the time we return to the outside we just want a path to a directory. Let me clean that up so it is clearer.

PR #174, opened 2023-10-04
chore: fix template for vector embedding

kmruiz:
## Checklist

### DIFF
--- a/packages/mongodb-constants/src/atlas-search-templates.ts
+++ b/packages/mongodb-constants/src/atlas-search-templates.ts
@@ -107,7 +107,9 @@ export const ATLAS_SEARCH_TEMPLATES: SearchTemplate[] = [
     "fields": {
       \${1:<vector embedding field name>}: [
         {
-          "type": "vectorEmbedding"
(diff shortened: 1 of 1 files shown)


packages/mongodb-constants/src/atlas-search-templates.ts
```diff
@@ -107,7 +107,9 @@ export const ATLAS_SEARCH_TEMPLATES: SearchTemplate[] = [
     "fields": {
       \${1:<vector embedding field name>}: [
         {
-          "type": "vectorEmbedding"
+          "type": "knnVector",
+          "dimensions": \${2:<number of dimensions>},
+          "similarity": \${3:<euclidean | cosine | dotProduct>}
```
Anemy, line 112: Would it make sense for the value for `similarity` be wrapped in a string? I'm wondering if that makes it easier for folks to use.
```suggestion
          "similarity": "\${3:<euclidean | cosine | dotProduct>}"
```

Anemy (approved):
One q, not a blocker.

PR #181, opened 2023-10-27
chore(sbom-tools): update Node.js version to test against

gribnoysup:
Node.js 16 is not supported anymore, so the tests are failing with "unsupported version of Node.js" error instead of trying to get the version CVE details that we test. Updated the version and the snapshots


addaleax (approved):

gribnoysup:
Oh, more tests to update the mocks for

PR #186, opened 2023-11-27
feat(mongodb-runner): add programmatic api for mongodb-runner

baileympearson:
## Description

I've been tinkering around with a vscode extension wrapper for mongodb-runner.  The package currently only exports the cluster helpers but does not provide a programmatic interface to the functionality contained in `cli.ts`.

I pulled logic for `ls`, `start` and `stop` into standalone, exported functions so that I can consume them in my vscode extension without running bash commands using the runner's CLI.  I understand if there's a desire to keep the API of this package slim and if we don't want to export the standalone helpers.  If that's the case, I would be okay with keeping them as standalone functions but not exporting them from the index.  I'm happy to import from the `runner-helpers.ts` file directly, this is just a toy project.

Along the way, I discovered that if I restarted my computer, `mongodb-runner ls` displays clusters that have been terminated.  I also added a `prune` command, both to the programmatic exports and the runner's CLI, that attempts to connect to every instance in the runner's `runnerDir` and removes all entries for clusters that cannot be connected to.

## Checklist


packages/mongodb-runner/src/util.ts
```diff
@@ -6,3 +6,16 @@ export const uuid = () => new BSON.UUID().toHexString(true);
...
   new Promise((r) => setTimeout(r, ms));
 export const range = (n: number): number[] => [...Array(n).keys()];
+export async function parallelForEach<T>(
+  iterable: AsyncIterable<T>,
+  fn: (arg0: T) => Promise<void> | void
+) {
+  const items = iterable[Symbol.asyncIterator]();
+
+  const result = [];
+  for (let item = await items.next(); !item.done; item = await items.next()) {
```
addaleax, line 16: Isn’t this just

```suggestion
  for await (const item of items) {
```

?

packages/mongodb-runner/src/util.ts
```diff
@@ -6,3 +6,16 @@ export const uuid = () => new BSON.UUID().toHexString(true);
...
+    result.push(fn(item.value));
+  }
+
+  return await Promise.allSettled(result);
```
addaleax, line 20: This effectively try/catches everything, is that intended here, especially since it’s a generic utility?

baileympearson: Yes.  The result is returned so it's accessable to callers, but this helper applies `fn` without waiting and awaits them all at the end.  The intention is to wait for all `fn`s to finish before returning, regardless of success or failure

I wrote this because `prune` was slow when there were many dead clusters, because `prune` waits `serverSelectionTimeoutMS` for each cluster.  I set `serverSelectionTimeoutMS` to 2s to speed `prune` up, but it's still slow when there are more than one or two clusters if each cluster is connected to sequentially

packages/mongodb-runner/src/util.ts
baileympearson: yeah :facepalm

packages/mongodb-runner/src/util.ts
addaleax: Yeah, totally valid use case and implementation :+1: I’d maybe consider incorporating something that calls out the lack of exception propagation to the function name or leave a comment about it, but the code itself is just fine

addaleax (approved):

PR #190, opened 2023-12-28
feat(signing-utils): add signing utils to devtools-shared

baileympearson:
## Description

Creates a new package to sign devtools artifacts with Garasign.  The package itself is a WIP - at a high level it provides a CLI interface for signing artifacts with Garasign.  It supports both local signing with docker, as well as signing over an SSH connection on remote that supports docker.

This PR also adds support for signing packages with jsign, although that has not been tested yet.

Other misc changes

- I encountered https://github.com/TypeStrong/ts-node/pull/2091 - so I bumped ts-node to latest

## Checklist


packages/signing-utils/src/signing-clients/index.ts
```diff
@@ -0,0 +1,62 @@
...
+import { LocalSigningClient } from './local-signing-client';
+import { RemoteSigningClient } from './remote-signing-client';
+
+export { LocalSigningClient } from './local-signing-client';
+export { RemoteSigningClient } from './remote-signing-client';
+
+export type SigningClientOptions = {
+  rootDir: string;
+  signingScript: string;
+  signingMethod: 'gpg' | 'jsign';
```
mabaasit, line 14: nit: we can have a type for this to avoid repetition.

mabaasit (approved):

packages/signing-utils/src/signing-clients/local-signing-client.ts
```diff
@@ -0,0 +1,53 @@
...
+
+      await this.copyFile(file, remotePath);
+      debug(`LocalSigningClient: Copied file ${file} to ${remotePath}`);
+
+      await execAsync(
+        `cd ${this.options.rootDir} && ./garasign.sh ${path.basename(file)}`,
+        {
+          env: {
+            method: this.options.signingMethod,
+          },
```
mabaasit, line 39: I just figured out that we need to pass the garasign and artifactory env vars here. i added [685329f](https://github.com/mongodb-js/compass/pull/5279/commits/685329feba42bd608b0868097ff12990572bb43b) in original pr

mcasimir (commented):
Looks super neat, left few comments, mostly NITs. Didn't press approve just yet, looks like there is something to fix based on [this](https://github.com/mongodb-js/devtools-shared/pull/190/files#r1439919219).

packages/signing-utils/src/mocha-hooks.ts
```diff
@@ -0,0 +1,11 @@
+import * as os from 'os';
+
+export const mochaHooks = {
+  beforeEach() {
+    // @ts-expect-error Stupid mocha binding properties to `this`
+    const test = this.currentTest ?? this.test;
+    if (os.platform().toLowerCase().includes('win32')) {
+      test.skip();
+    }
```
gribnoysup, line 9: I was pretty sure that just `this.skip` is a thing in beforeEach and you don't need to specifically get the spec first, just wondering if this expect-error is needed 🤔

packages/signing-utils/bin/sign.js
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
+const { sign } = require('./../dist');
```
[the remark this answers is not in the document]
mabaasit: for Compass, the `wininstaller` `signtool` which we patch, we need to built a binary to replace it and we need this cli there.

packages/signing-utils/src/garasign.sh
```diff
@@ -0,0 +1,67 @@
...
+ARTIFACTORY_HOST="artifactory.corp.mongodb.com"
+
+logout_artifactory() {
+  docker logout "${ARTIFACTORY_HOST}" > /dev/null 2>&1
+  echo "Logged out from artifactory"
+}
+trap logout_artifactory EXIT
+
+echo "Logging into docker artifactory"
+echo "${artifactory_password}" | docker login --password-stdin --username ${artifactory_username} ${ARTIFACTORY_HOST} > /dev/null 2>&1
```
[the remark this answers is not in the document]
baileympearson: we do need to sign in - we have special artificatory credentials for our evergreen projects. I am unsure about signing out. for our evergreen projects. I am unsure about signing out. 
I think it's probably best practice to always sign out.

Good catch signing in on windows--I haven't tested the script on windows yet, so I missed that.

packages/signing-utils/src/garasign.sh
```diff
@@ -0,0 +1,67 @@
...
+EOL
+
+directory=$(pwd)
+file=$1
+
+echo "File to be signed: $file"
+echo "Working directory: $directory"
+
+gpg_sign() {
+  docker run \
```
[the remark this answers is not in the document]
baileympearson: just a miss - good catch

packages/signing-utils/src/signing-clients/index.ts
baileympearson: fixed

packages/signing-utils/src/signing-clients/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  options: ClientOptions
+): Promise<SigningClient> {
+  async function getSshClient(sshOptions: ConnectConfig) {
+    const sshClient = new SSHClient(sshOptions);
+    await sshClient.connect();
+    return sshClient;
+  }
+
+  function getSigningScript() {
+    return path.join(__dirname, '..', 'src', './garasign.sh');
```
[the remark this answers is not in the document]
baileympearson: Fixed

packages/signing-utils/src/signing-clients/local-signing-client.ts
```diff
@@ -0,0 +1,53 @@
+import path from 'path';
+import { exec } from 'child_process';
```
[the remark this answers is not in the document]
baileympearson: yep, I changed the script to use execFile

packages/signing-utils/src/signing-clients/local-signing-client.ts
baileympearson: I believe I address this comment, the generic helper will gather all environment variables and I assign them here. Let me know if I missed something.

packages/signing-utils/src/utils.ts
```diff
@@ -0,0 +1,16 @@
+import { debug as debugFn } from 'debug';
+
+export const debug = debugFn('signing-utils');
+
+export function assertRequiredVars() {
```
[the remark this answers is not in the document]
baileympearson: I think this might no longer be relevant-I changed the file to gather environment variables needed for code signing. It is now used in multiple locations.

packages/signing-utils/src/mocha-hooks.ts
baileympearson: Because the hooks are declared in a regular object, typescript is unable to infer that a skip method exists on this object. This is still true, even if I explicitly type beforeEach.

I did make your suggested change though, we just still need the expect error.

packages/signing-utils/src/signing-clients/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    return sshClient;
+  }
+
+  const signingScript = path.join(__dirname, '../..', 'src', './garasign.sh');
+
+  if (options.client === 'remote') {
+    const sshClient = await getSshClient(options);
+    // Currently only linux remote is supported to sign the artifacts
+    return new RemoteSigningClient(sshClient, {
+      workingDirectory: '~/garasign',
```
mabaasit, line 66: I know this path value is here from my PR, but after debugging for some time, I realized this path is not expanded by default when copying files to the remote server. The sftp client interprets it exactly as this and throws error `No such file`. We need to expand it here to `/home/ubuntu/garasign` or some other absolute path.

packages/signing-utils/src/cli.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const file = program.args[0];
+
+sign(file, program.opts()).then(
+  () => {
+    // do nothing
+  },
+  (error) => {
+    // eslint-disable-next-line no-console
+    console.error({ error });
```
addaleax, line 46:
```suggestion
    console.error({ error });
    process.exitCode = 1;
```

packages/signing-utils/src/garasign.sh
```diff
@@ -0,0 +1,62 @@
...
+    -e GRS_CONFIG_USER1_USERNAME="${garasign_username}" \
+    -e GRS_CONFIG_USER1_PASSWORD="${garasign_password}" \
+    --rm \
+    -v $directory:$directory \
+    -w $directory \
+    ${ARTIFACTORY_HOST}/release-tools-container-registry-local/garasign-gpg \
+    /bin/bash -c "gpgloader && gpg --yes -v --armor -o $file.sig --detach-sign $file"
```
addaleax, line 42:
```suggestion
    /bin/bash -c "gpgloader && gpg --yes -v --armor -o '$file.sig' --detach-sign '$file'"
```

(let's try to make this somewhat resistant against bash expansions from arbitrary filenames?)

packages/signing-utils/src/garasign.sh
```diff
@@ -0,0 +1,62 @@
...
+
+jsign_sign() {
+  docker run \
+    -e GRS_CONFIG_USER1_USERNAME="${garasign_username}" \
+    -e GRS_CONFIG_USER1_PASSWORD="${garasign_password}" \
+    --rm \
+    -v $directory:$directory \
+    -w $directory \
+    artifactory.corp.mongodb.com/release-tools-container-registry-local/garasign-jsign \
+    /bin/bash -c "jsign --tsaurl "timestamp.url" -a mongo-authenticode-2021 $file"
```
addaleax, line 53:
```suggestion
    /bin/bash -c "jsign --tsaurl "timestamp.url" -a mongo-authenticode-2021 '$file'"
```

packages/signing-utils/src/index.spec.ts
addaleax, line 1: Delete this file?

packages/signing-utils/src/signing-clients/remote-signing-client.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  private async signRemoteFile(file: string) {
+    const env = getEnv();
+    /**
+     * Passing env variables as an option to ssh.exec() doesn't work as ssh config
+     * (`sshd_config.AllowEnv`) does not allow to pass env variables by default.
+     * So, here we are passing the env variables as part of the command.
+     */
+    const cmds = [
+      `cd ${this.options.workingDirectory}`,
```
addaleax, line 80:
```suggestion
      `cd '${this.options.workingDirectory}'`,
```

packages/signing-utils/src/signing-clients/remote-signing-client.ts
```diff
@@ -0,0 +1,117 @@
...
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      `export garasign_username=${env.garasign_username}`,
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      `export garasign_password=${env.garasign_password}`,
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      `export artifactory_username=${env.artifactory_username}`,
+      // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+      `export artifactory_password=${env.artifactory_password}`,
+      `export method=${this.options.signingMethod}`,
+      `./garasign.sh ${file}`,
```
addaleax, line 90:
```suggestion
      `./garasign.sh '${file}'`,
```

packages/signing-utils/src/mocha-hooks.ts
gribnoysup: Ohhh, right, because you're not in an actual test file calling a global beforeEach, totally missed that, makes sense!

packages/signing-utils/bin/sign.js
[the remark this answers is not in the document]
mcasimir: we need to built a binary to replace it and we need this cli there.

Why would we need a CLI though? not strongly opposed to it, but it doesn't feel strictly necessary if we need it only for wininstaller in Compass, we can just have `node ./hadron-build/signtool.js`, and in `./hadron-build/signtool.js` require this package and run sign, so we have one less interface to maintain.

packages/signing-utils/src/signing-clients/index.ts
baileympearson: Could we use [realpath](https://github.com/mscdex/ssh2/blob/master/SFTP.md#:~:text=Error%20%3Eerr.-,realpath,-(%3C%20string%20%3Epath) to expand the working directory before we start using the RemoteClient?

baileympearson: Seems like maybe an integration test would be valuable here.  What if we use ssh2 to spin up a testing ssh server to e2e test the remote client?  The only mechanism not tested in that case is the actual signing with Garasign.

packages/signing-utils/src/signing-clients/index.ts
```diff
@@ -0,0 +1,82 @@
...
+export async function getSigningClient(
+  options: ClientOptions
+): Promise<SigningClient> {
+  async function getSshClient(sshOptions: ConnectConfig) {
+    const sshClient = new SSHClient(sshOptions);
+    await sshClient.connect();
+    return sshClient;
+  }
+
+  const signingScript = path.join(__dirname, '../..', 'src', './garasign.sh');
```
mabaasit, line 62: Ignore if i'm wrong about this, but would this work on windows?

addaleax: yes, `/` works as a path separator on Windows just like `\` does :) so this could for example also be written as

```suggestion
  const signingScript = path.join(__dirname, '..', '..', 'src', 'garasign.sh');
```

or 

```suggestion
  const signingScript = path.join(__dirname, '../../src/garasign.sh');
```

and those would return the same result

packages/signing-utils/src/ssh-client.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+  });
+
+  it('connects successfully on ready', async function () {
+    const connectStub = sandbox
+      .stub(sshClient['sshConnection'], 'connect')
+      .resolves(sshClient['sshConnection']);
+    await Promise.all([
+      sshClient.connect(),
+      sshClient['sshConnection'].emit('ready'),
+    ]);
```
addaleax, line 30: Does `.emit('ready')` actually return a Promise? I can't find anything to imply that it does

If the goal here is to execute code while the Promise returned from `.connect()` is pending, I'd just do that in the explicit (and shorter) version, i.e.

```suggestion
    const connectPromise = sshClient.connect();
    sshClient['sshConnection'].emit('ready');
    await connectPromise;
```

(ditto below)

packages/signing-utils/src/ssh-client.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (!this.connected) {
+      throw new Error('Not connected to ssh server');
+    }
+    return new Promise((resolve, reject) => {
+      this.sshConnection.exec(command, (err, stream) => {
+        if (err) {
+          return reject(err);
+        }
+        let data = '';
+        stream.on('data', (chunk: string) => {
```
addaleax, line 66:
```suggestion
        stream.setEncoding('utf8').on('data', (chunk: string) => {
```

(otherwise you'll into problems with multibyte characters split across chunks and the parameter type signature would be `chunk: Buffer`, not `chunk: string`)

packages/signing-utils/src/ssh-client.ts
```diff
@@ -0,0 +1,106 @@
...
+          data += chunk;
+        });
+        stream.stderr.on('data', (chunk) => {
```
addaleax, line 69:
```suggestion
        stream.stderr.setEncoding('utf8').on('data', (chunk) => {
```

packages/signing-utils/src/ssh-client.ts
```diff
@@ -0,0 +1,106 @@
...
+      ? await readFile(this.sshClientOptions.privateKey)
+      : undefined;
+    return new Promise((resolve, reject) => {
+      this.sshConnection.connect({
+        ...this.sshClientOptions,
+        privateKey,
+      });
+      this.sshConnection.on('error', reject);
+      // @ts-expect-error We expect an error here - why?
+      this.sshConnection.on('ready', resolve);
```
addaleax, line 47: The "why" for this error is because TS cannot infer the type of the returned Promise here, so it falls back to `Promise<unknown>` and gives `resolve` a signature of `(value: unknown) => void` (i.e. one non-optional argument), which doesn't match the expected type for the callback passed to `.on('ready', cb)` (i.e. a function taking zero non-optional arguments), but since there are other events declared whose callbacks _do_ take arguments, TS gives a non-helpful error message thinking you passed the wrong event name.

There are a couple ways to go about this, the (imo) best one being to tell TS that you're returning a `Promise<void>` here; There is already a linter warning on this method and others about missing return types, just fixing those should also address this particular issue.

baileympearson: That makes sense. Thanks for the explanation! I was really scratching my head as to why Typescript couldn't find the right overload 😁

packages/signing-utils/src/ssh-client.ts
baileympearson: This is interesting, not something I've ever considered. Thanks for sharing!

I made the change, suggested here, but just because I'm curious – would leaving the encoding unspecified, concatenating buffers in each callback, and then stringifying the result at the end achieve the same result?  I'm not sure if that would be more performance, because I don't know of a way to concatenate buffers without copying the original. In the driver, we maintain a custom data structure for this purpose.

packages/signing-utils/src/ssh-client.spec.ts
baileympearson: Yeah, good catch. I didn't look that closely at this test file, I ended up re-working it.

packages/signing-utils/src/ssh-client.ts
addaleax: Yes, gathering all chunks as Buffers and then concatenating them would give the same result – I'd expect it to be significantly slower, though, there's a lot more objects and crossings between JS and C++ involved in that case :)

addaleax (approved):

mcasimir (approved):

PR #225, opened 2024-02-05
fix(oidc-mock-provider): move skipIdToken to callback response

paula-stacho:
## Description
Actually, what I did before made no sense, because `getTokenPayload` is defined out of the plugin and called from within 🤦 
https://github.com/mongodb-js/oidc-plugin/pull/187/files


addaleax (approved):
Sorry I also totally missed this 😅

PR #239, opened 2024-02-16
chore: import from https://github.com/mongodb-js/devtools-connect COMPASS-7645

paula-stacho:
## Description
Not a straight-forward copy-paste. But still the diff is too big, so let me tell you what has been done.

### oidc-http-server-pages
New package was created with `npm run create-workspace`. 
The contents (src) were taken from https://github.com/mongodb-js/devtools-connect/tree/main/src/oidc/static.
New `index.ts` was added which exposes all the exports.
The rest of adjustions was to make sure the npm scripts are working. Files were reformatted to fit the new prettier and linting, and selected dependencies from `devtools-connect` were added. The compilation follows devtools-connect in that it includes `static-pages.js` and `.esm-wrapper.mjs` (this part I'm not sure if it's necessary). 

### devtools-connect
Again, new package was created with `npm run create-workspace`.
The contents (src, scripts & package.json) were taken from https://github.com/mongodb-js/devtools-connect. Usages of `oidc-http-server-pages` were replaced with the new separate package.
Again, I had to ensure the npm scripts are working. For this, I added the aliases used by lerna. Again, files were reformatted to fit new prettier. Linting was more difficult and I ended up just using the old config from `devtools-connect` + turning off a couple of extra rules that were in conflict with prettier.

Lastly, the packages were tested by linking the devtools-connect package to mongosh and running `npm run test` and `npm run test-e2e`.


packages/devtools-connect/package.json
```diff
@@ -0,0 +1,94 @@
...
+    "prepack": "npm run compile",
+    "compile-ts": "tsc -p tsconfig.json"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@mongodb-js/oidc-http-server-pages": "0.1.0",
+    "lodash.merge": "^4.6.2",
+    "mongodb-connection-string-url": "^3.0.0",
+    "system-ca": "^1.0.2",
+    "mongodb": "^5.8.1 || ^6.0.0"
```
addaleax, line 54: You're running into the problem with `saslprep`, right? You can just add saslprep itself as a `devDependencies` entry here to solve that (we probably don't want to directly make `mongodb` a dependency of `devtools-connect`, just a peer dependency like it previously was)

paula-stacho: oh, thank you!

packages/oidc-http-server-pages/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export { generateStaticPagesModule } from './create-static-pages';
```
addaleax, line 1: You could probably drop this export, `./create-static-pages` isn't something that consumers of this package should actually need

packages/devtools-connect/src/fast-failure-connect.ts
```diff
@@ -0,0 +1,13 @@
+// It probably makes sense to put this into its own package/repository once
+// other tools start using it.
+
+export function isFastFailureConnectionError(error: Error): boolean {
+  switch (error.name) {
+    case 'MongoNetworkError':
+      return /\b(ECONNREFUSED|ENOTFOUND|ENETUNREACH)\b/.test(error.message);
```
addaleax, line 7:
```suggestion
      return /\b(ECONNREFUSED|ENOTFOUND|ENETUNREACH|EINVAL)\b/.test(error.message);
```

might as well address https://jira.mongodb.org/browse/MONGOSH-1717 while we're already in the process of making changes 😄 

(but feel free to ignore this if it's easier 🙂)

packages/devtools-connect/package.json
```diff
@@ -0,0 +1,94 @@
...
+    "ts-node": "^10.2.1",
+    "ts-sinon": "^2.0.1",
+    "typescript": "^5.0.4"
+  },
+  "optionalDependencies": {
+    "kerberos": "^2.1.0",
+    "mongodb-client-encryption": "^6.0.0",
+    "os-dns-native": "^1.2.0",
+    "resolve-mongodb-srv": "^1.1.1",
+    "socks": "^2.7.3"
```
addaleax, line 92: There are a few dependencies of the driver listed here: https://github.com/mongodb-js/devtools-connect/blob/34c8e2da3a0296c269ba8f8af44fbee0fab76549/package.json#L80-L83

Do you remember why these are being added here? I don't think it's a bad thing, just wanted to check.

(Something like `socks` could probably just be a regular dependency, it's not a native addon, it's just an optional feature of the driver but one that we want to support in all of our products)

addaleax (approved):
I know this is still labeled Draft, but everything looks great! 🚀

packages/devtools-connect/.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  root: true,
+  extends: [
+    'semistandard',
+    'plugin:promise/recommended',
+    'plugin:@typescript-eslint/recommended',
+  ],
+  rules: {
```
mcasimir, line 8: is fine for a follow up, but generally we try to avoid doing this, and instead we align on eslint / ts when we move things here, that way we can keep maintaining everything in the same way, especially for stuff that can be just addressed with `eslint --fix`

saisrinivas795 (approved):

packages/devtools-connect/package.json
paula-stacho: ah right! so devtools-connect didn't have a depcheck, and now that it was added, there were several dependencies marked missing. I added them as optional as it seemed the package was doing fine without them.

paula-stacho: I can move the socks, mongodb-client-encryption and kerberos are fine like this?

packages/devtools-connect/.eslintrc.js
paula-stacho: yeah, this will have to be a follow up 🙈 . I mention this in the description - I have tried to make this work with the same rulesets as the other packages, but I kept seeing errors that I couldn't even figure out where they were coming from. so in the end I gave up and used the old eslint config + a couple of exceptions to support the new prettier config.

packages/devtools-connect/src/fast-failure-connect.ts
paula-stacho: sure! I was able to reproduce this with https://github.com/mongodb-js/mongosh/pull/1832, we can add this test once devtools-connect is updated

packages/devtools-connect/package.json
addaleax: Hm, I guess the depcheck here in the repo picks up on the dynamic `require()` calls 🙂 Yes, definitely fine this way

mcasimir (approved):

PR #246, opened 2024-02-26
feat(oidc-http-server-pages): generate JSON templates COMPASS-7646

paula-stacho:
## Checklist


packages/oidc-http-server-pages/src/create-templates.ts
```diff
@@ -0,0 +1,111 @@
...
+}: {
+  Component: Component;
+  parameters: string[];
+}): ITemplate[] {
+  const templates: ITemplate[] = [];
+  for (const paramsSubset of allSubsets(parameters)) {
+    const propsObject = paramsSubset.reduce((obj, prop) => {
+      obj[prop] = placeholder(prop);
+      return obj;
+    }, {} as Record<string, string>);
```
addaleax, line 55: This is the same as doing

```suggestion
    const propsObject = Object.fromEntries(
      paramsSubset.map((prop) => [prop, placeholder(prop)])
    );
```

right? No strong feelings, but I personally find that a bit simpler to read

paula-stacho: yes it's the same, sure!

packages/oidc-http-server-pages/src/get-static-page.ts
```diff
@@ -0,0 +1,61 @@
...
+  parameters: Record<string, string | undefined>
+) {
+  let { html } = template;
+  for (const [key, placeholder] of Object.entries(template.parameters)) {
+    html = html.replaceAll(placeholder, escapeHTML(parameters[key] as string));
+  }
+  return html;
+}
+
+function escapeHTML(str: string) {
```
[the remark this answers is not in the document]
paula-stacho: is it a best practice to use explicit return types always? I thought with simple ones like this there is no need. a quick search doesn't tell me otherwise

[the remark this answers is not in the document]
addaleax: Yeah, I'd say so – I thought we had https://typescript-eslint.io/rules/explicit-function-return-type/ enabled in a few places, but I guess it didn't make it into the "shared" eslint config in this repo. I'd generally recommend always doing this, yes, unless the type is hard to spell out and inconsequential, otherwise it's a good idea to check that you return the type that you're actually expecting to return

packages/oidc-http-server-pages/src/create-static-pages.ts
```diff
@@ -1,146 +0,0 @@
...
-          'errorDescription',
-          'errorURI',
-          'productDocsLink',
-          'productName',
-        ],
-      ],
-      [OIDCAcceptedPage, ['productDocsLink', 'productName']],
-      [OIDCNotFoundPage, ['productDocsLink', 'productName']],
-    ])
-  );
```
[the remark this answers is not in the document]
paula-stacho: Makes sense, I will look into this. Thank you!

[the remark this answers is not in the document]
paula-stacho: So, now we're creating two files -> one .gz and one .js that decompresses and exports the templates. I kept brotli for the latter, the file is now 12 KB. I hope it's okay that the files are created like this during the build. Previously I kept the "console.log solution" because I assumed there was a reason for it, such as permissions when the build happens in another context. Do you know if that was the case or not? @addaleax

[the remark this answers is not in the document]
paula-stacho: Also I kept the brotli settings, but I wonder if the MAX_QUALITY setting makes sense if we already use the MODE_TEXT. I don't know much about compression, but it seems weird that for text compression there would be different quality settings 🤔

packages/oidc-http-server-pages/src/create-templates.ts
```diff
@@ -84,26 +84,23 @@ export function generateTemplates<TPage extends string = HttpServerPage>(
...
+      [zlibConstants.BROTLI_PARAM_QUALITY]: zlibConstants.BROTLI_MAX_QUALITY,
+    },
+  });
   writeFileSync(
     path.join(__dirname, 'templates.js'),
     `
-    const { gunzipSync } = require('zlib');
-    const buffer = gunzipSync(
+    const { brotliDecompressSync } = require('zlib');
+    const buffer = brotliDecompressSync(
```
addaleax, line 103: I think ideally we'd also keep the lazy decompression that the previous code had – otherwise, loading this file would always uncompress the data blob, regardless of whether we'll actually end up using it.

(I'll probably do something similar for the devtools-connect package in general as part of the shell performance work for loading `oidc-plugin` from `devtools-connect`, it turns out instantiating the npm packages used for the HTTP server is fairly expensive, but I'm not sure about what exactly that will look like)

packages/oidc-http-server-pages/src/create-static-pages.ts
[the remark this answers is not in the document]
addaleax: I think they're fairly independent settings – MODE_TEXT is just a hint to the compressor that it's compressing an UTF-8-encoded string and not arbitrary binary data, MAX_QUALITY means that it's okay to be fairly slow while compressing data in exchange for a smaller compressed output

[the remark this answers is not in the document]
paula-stacho: ah, I misunderstood what MAX_QUALITY means then

packages/oidc-http-server-pages/src/create-templates.ts
paula-stacho: I thought the conditional require in getStaticPage would do the job, is that not so?

paula-stacho: ah, right it will go through webpack

addaleax: Ah no right, it should address this 👍

paula-stacho: oh okay, I was too quick and I realised I'm not sure if conditional require even works with webpack

addaleax: @paula-stacho Yeah, the only difference with webpack is that the code will be _parsed_ regardless of whether it is used or not, but conditional _evaluation_ of modules works just fine

paula-stacho: yeah that makes sense. I will revert the last commits.

PR #254, opened 2024-03-06
fix(saslprep): saslprep typing in node.d.ts NODE-5996

baileympearson:
## Description

the recent changes to make saslprep compatible with the browser changed the generated types for Nodejs.  When we tried to consume the latest saslprep in the driver, we get compile errors.

It seems like the combination of `_saslprep.bind` and `Object.assign` erase the type of `saslprep` such that the generated .d.ts file doesn't accurately reflect the types.  

`node.d.ts` before the changes in this PR:
```typescript
declare const saslprep: (args_0: string, args_1?: {
    allowUnassigned?: boolean | undefined;
} | undefined) => string;
export = saslprep;
//# sourceMappingURL=node.d.ts.map
```

`node.d.ts` with the changes in this PR:
```typescript
declare function saslprep(input: string, opts?: {
    allowUnassigned?: boolean;
}): string;
declare namespace saslprep {
    export var saslprep: typeof import("./node");
    var _a: typeof import("./node");
    export { _a as default };
}
export = saslprep;
//# sourceMappingURL=node.d.ts.map
```

## Checklist

### DIFF
--- a/packages/saslprep/src/node.ts
+++ b/packages/saslprep/src/node.ts
@@ -4,8 +4,11 @@ import data from './code-points-data';
 
 const codePoints = createMemoryCodePoints(data);
 
-const saslprep = _saslprep.bind(null, codePoints);
+function saslprep(input: string, opts?: { allowUnassigned?: boolean }): string {
+  return _saslprep(codePoints, input, opts);
+}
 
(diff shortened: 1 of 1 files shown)


addaleax (approved):

nbbeeken (approved):
LGTM

PR #281, opened 2024-04-08
Revert "chore(saslprep): skip codepoint generation in CI (#188)"

baileympearson:
This reverts commit c29f69bfd842e6009861ddf8d2e5542797e83423.

## Checklist

### DIFF
--- a/packages/saslprep/src/generate-code-points.ts
+++ b/packages/saslprep/src/generate-code-points.ts
@@ -41,7 +41,8 @@ memory.push(
   traverse(bidirectional_l, codePoints.bidirectional_l)
 );
 
-process.stdout.write(
+const fsStream = createWriteStream(process.argv[2]);
+fsStream.write(
   `import { gunzipSync } from 'zlib';
 
 export default gunzipSync(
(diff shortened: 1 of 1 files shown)


addaleax (approved):

packages/saslprep/src/generate-code-points.ts
```diff
@@ -1,13 +1,6 @@
 import { gzipSync } from 'zlib';
 import bitfield from 'sparse-bitfield';
 import * as codePoints from './code-points-src';
-import { createWriteStream } from 'fs';
-
-if (!process.env.GENERATE_CODE_POINTS) {
-  // TODO(COMPASS-7555): fix codepoint generation=
-  process.exitCode = 0;
-  process.exit();
-}
```
addaleax, line 10: I don't think we can easily remove the `createWriteStream` addition again because now this script has 2 output files

But we could also just remove this conditional as a revert, I think

packages/saslprep/src/generate-code-points.ts
```diff
@@ -47,8 +41,7 @@ memory.push(
   traverse(bidirectional_l, codePoints.bidirectional_l)
 );
 
-const fsStream = createWriteStream(process.argv[2]);
-fsStream.write(
+process.stdout.write(
```
addaleax, line 44: This would also need to keep using `createWriteStream` (or otherwise the package.json bits also need to come back, but I think it might be weird to create one file by writing to stdout and one by writing to disk)

addaleax (approved):

PR #296, opened 2024-04-24
feat: add azure to the ADL $out description & comment COMPASS-7419

lerouxb:
https://jira.mongodb.org/browse/COMPASS-7419

The suggested snippet seemed to have it as the new snippet either in addition to or in place of s3, but our established pattern seems to be to have one option in the snippet and the others documented in the comment so I stuck with that.

### DIFF
--- a/packages/mongodb-constants/src/stage-operators.ts
+++ b/packages/mongodb-constants/src/stage-operators.ts
@@ -681,7 +681,7 @@ const STAGE_OPERATORS = [
     apiVersions: [1],
     namespaces: [...ANY_NAMESPACE],
     description:
-      'Writes the result of a pipeline to an Atlas cluster or S3 bucket.',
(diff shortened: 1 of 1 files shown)


packages/mongodb-constants/src/stage-operators.ts
```diff
@@ -681,7 +681,7 @@ const STAGE_OPERATORS = [
     apiVersions: [1],
     namespaces: [...ANY_NAMESPACE],
     description:
-      'Writes the result of a pipeline to an Atlas cluster or S3 bucket.',
+      'Writes the result of a pipeline to an Atlas cluster, S3 bucket or Azure Blob Storage.',
```
Anemy, line 684: nit, serial comma
```suggestion
      'Writes the result of a pipeline to an Atlas cluster, S3 bucket, or Azure Blob Storage.',
```

Anemy (approved):
lgtm, one text suggestion, also double checking, did you chat with Irwin on the syntax mentioned in the ticket?

PR #321, opened 2024-05-15
fix(devtools-connect): give application code for custom OIDC HTTP options precedence

addaleax:
Instead of using the application callback first and then potentially transforming the result with options computed by devtools-connect, do the reverse. This gives the application using devtools-connect more flexibility about how exactly to modify OIDC HTTP options.

The goal here is to enable usage of HTTP proxies in conjunction with `useSystemCA` and explicit `tlsCAFile` settings; devtools-connect correctly passes the former to the OIDC plugin, but after this change, the application can also pass it to the proxy agent, if it creates one.

None of our team's applications currently make use of the function variant of this options, so this will not affect existing code in any way.

### DIFF
--- a/packages/devtools-connect/package.json
+++ b/packages/devtools-connect/package.json
@@ -56,7 +56,7 @@
   "peerDependencies": {
     "@mongodb-js/oidc-plugin": "^0.4.0",
     "mongodb": "^5.8.1 || ^6.0.0",
-    "mongodb-log-writer": "^1.2.0"
+    "mongodb-log-writer": "^1.4.2"
   },
   "devDependencies": {
     "@mongodb-js/oidc-plugin": "^0.4.0",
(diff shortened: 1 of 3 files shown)


packages/devtools-connect/package.json
```diff
@@ -76,7 +76,7 @@
     "gen-esm-wrapper": "^1.1.0",
     "mocha": "^8.4.0",
     "mongodb": "^5.8.1 || ^6.0.0",
-    "mongodb-log-writer": "^1.2.0",
+    "mongodb-log-writer": "^1.4.2",
```
addaleax, line 79: I don't even quite remember why, it must have been part of the SSDLC CodeQL work, but this was lying around in my local git checkout and there didn't really seem a reason not to just do this as a drive-by change, so I just commited it along with the actual change 🤷‍♀️

PR #353, opened 2024-06-06
fix(query-parser): provide stringify handler for BSONRegExp COMPASS-7008

Anemy:
COMPASS-7008

We were providing a stringify option for the loose javascript representation of RegExp, but not the BSONRegExp which is what we serialize into in Compass.


lerouxb (approved):

packages/query-parser/src/stringify.ts
```diff
@@ -134,6 +135,12 @@ const BSON_TO_JS_STRING = {
 
     return `RegExp(${JSON.stringify(v.source)}${hasOptions ? `, '${o}'` : ''})`;
   },
+  BSONRegExp: function (v: BSONRegExp) {
+    const hasOptions = v.options && v.options?.length > 0;
+    return `RegExp(${JSON.stringify(v.pattern)}${
+      hasOptions ? `, '${v.options}'` : ''
+    })`;
```
addaleax, line 142: Do we want to account for regular expressions that cannot be represented by JS RegExps here as well? e.g.

```suggestion
    let ctor = 'RegExp';
    try {
      RegExp(v.pattern, v.options || undefined);
    } catch {
      ctor = 'BSONRegExp';
    } 
    return `${ctor}(${JSON.stringify(v.pattern)}${
      hasOptions ? `, '${v.options}'` : ''
    })`;
```

Anemy: Nice catch, adding a test for this now.

Anemy: Added 2 tests. 
TIL: https://www.mongodb.com/docs/manual/reference/operator/query/regex/#pcre-versus-javascript 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions
https://www.mongodb.com/docs/manual/reference/operator/query/regex/

One design decision here is that we're using BSONRegExp and not the $regex/$options operator. Something we could change down the line I suppose.

addaleax (approved):

PR #390, opened 2024-07-08
feat(mongodb-redact): move package into monorepo

addaleax:
Separated into commits for easier review:

##### fixup: rename package to old name

### DIFF
--- a/packages/mongodb-redact/src/index.ts
+++ b/packages/mongodb-redact/src/index.ts
@@ -1,26 +1,26 @@
-var _ = require('lodash');
-var regexes = require('./regexes');
+import _ from 'lodash';
+import { regexes } from './regexes';
 
-function redact(message) {
+export function redact<T>(message: T): T {
   if (_.isPlainObject(message)) {
     // recursively walk through all values of an object
-    return _.mapValues(message, redact);
+    return _.mapValues(message as any, redact);
   }
   if (_.isArray(message)) {
     // walk through array and redact each value
-    return _.map(message, redact);
+    return _.map(message, redact) as T;
   }
-  if (!_.isString(message)) {
+  if (typeof message !== 'string') {
     // all non-string types can be safely returned
     return message;
   }
   // apply all available regexes to the string
-  _.each(regexes, function (values) {
-    var regex = values[0];
-    var replacement = values[1];
-    message = message.replace(regex, replacement);
-  });
+  for (const [regex, replacement] of regexes) {
+    // The type here isn't completely accurate in case `T` is a specific string template
+    // but it's close enough for practical usage
+    message = (message as T & string).replace(regex, replacement) as T & string;
+  }
   return message;
 }
 
-module.exports = redact;
+export default redact;
--- a/packages/mongodb-redact/src/index.ts
+++ b/packages/mongodb-redact/src/index.ts
@@ -1,0 +1,26 @@
+var _ = require('lodash');
+var regexes = require('./regexes');
+
+function redact(message) {
+  if (_.isPlainObject(message)) {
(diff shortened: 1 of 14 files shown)


packages/mongodb-redact/src/index.ts
```diff
@@ -1,26 +1,26 @@
...
-    return _.mapValues(message, redact);
+    return _.mapValues(message as any, redact);
   }
   if (_.isArray(message)) {
     // walk through array and redact each value
-    return _.map(message, redact);
+    return _.map(message, redact) as T;
   }
-  if (!_.isString(message)) {
+  if (typeof message !== 'string') {
```
Anemy, line 13: The lodash type check here also allows for an object string type: 
https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L12135
Is that something we should care about? I'm thinking fine as is, but I figured it's worth bringing up.
```
assert.strictEqual(_.isString(Object('a')), true);
```

packages/mongodb-redact/src/regexes.ts
```diff
@@ -38,18 +38,16 @@ var regexes = [
...
   [
-    /(http(s)?:\/\/)(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}(\.[a-z]{2,6})?\b([-a-zA-Z0-9@:%_\+.~#?&\/\/=]*)/gim,
+    /(http(s)?:\/\/)(www\.)?[-a-zA-Z0-9@:%._+~#=]{2,256}(\.[a-z]{2,6})?\b([-a-zA-Z0-9@:%_+.~#?&/=]*)/gim,
     '<url>',
   ],
 
   // MongoDB connection strings
   [
-    /(mongodb:\/\/)(www\.)?[-a-zA-Z0-9@:%._\+~#=,]{2,256}(\.[a-z]{2,6})?\b([-a-zA-Z0-9@:%_\+.~#?&\/\/=]*)/gim,
+    /(mongodb:\/\/)(www\.)?[-a-zA-Z0-9@:%._+~#=,]{2,256}(\.[a-z]{2,6})?\b([-a-zA-Z0-9@:%_+.~#?&/=]*)/gim,
```
Anemy, line 47: The end part here `[-a-zA-Z0-9@:%_+.~#?&/=]` used to contain a `\/` character`[-a-zA-Z0-9@:%_\+.~#?&\/\/=]` . Is that not needed anymore?

Anemy (approved):
lgtm! Two small questions on some nuanced behavioral changes which likely shouldn't impact things, not blockers.

packages/mongodb-redact/src/regexes.ts
addaleax: These are all automated replacements, no manual changes – `\/` inside a `[]` is equivalent to just `/` and `//` inside `[]` is equivalent to `/`.

addaleax: (This did make me wonder if there was an intention to target a different character, since listing `/` twice seems odd. But I wanted to avoid functional changes in this PR, it's just for moving things to monorepo/TypeScript)

packages/mongodb-redact/src/index.ts
addaleax: That's honestly pretty surprising to me... I guess technically `.replace()` would still work even on those, but tbh, I'd assume that the only reason this was using `_.isString()` was that whoever wrote this just used lodash for everything and didn't think much of this?

Anemy: Yeah I'd prefer we don't use lodash at all, but I think that's not something for now. All good!

PR #393, opened 2024-07-09
feat!(ejson-shell-parser): import into monorepo

addaleax:
This is marked `feat!()` because it does not copy over the build tooling from the original repo, so will result in different output (e.g. no bundle creation).

##### feat!(ejson-shell-parser): create package

Worth pointing out that the original package did not include a license, but had a "MIT" license tag in its package.json. Since contributors have to sign the CLA, standardizing on Apache 2.0 here should be fine and is the right thing to do in general.


packages/ejson-shell-parser/src/eval.ts
```diff
@@ -98,7 +103,7 @@ const memberExpression = (node: CallExpression, withNew: boolean): any => {
         throw new Error('Expected CallExpression property to be an identifier');
 
       const args = node.arguments.map((arg) => walk(arg));
-      return calleeThis[calleeFn].apply(calleeThis, args);
+      return calleeThis[calleeFn](...args);
```
Anemy, line 106: Just checking, none of these `calleeThis` functions use `this` in their own scope right?

Anemy (approved):
lgtm, one question for my understanding. Nice to have this in this repo. In the slack thread I asked about the possible renaming of this package, not sure if it's something we want to do yet though.

PR #401, opened 2024-07-10
fix(redact): use fast and safe path and email regex MONGOSH-1392

alenakhineika:
## Description

- Replace user name in user path
- Do not modify the electron path only hide a user name
- Make all URLs pass the redos checker

## Checklist

### DIFF
--- a/packages/mongodb-redact/src/regexes.ts
+++ b/packages/mongodb-redact/src/regexes.ts
@@ -13,11 +13,11 @@ export const regexes = [
 
   // User directories
   [
-    /(file:\/\/|\/)(Users|user|users|user|usr|home|u01|var\/users|export|Documents and Settings|Profiles)\/[^/]*\//gm,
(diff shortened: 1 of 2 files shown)


packages/mongodb-redact/src/regexes.ts
```diff
@@ -11,10 +11,21 @@ export const regexes = [
...
+  ],
+
+  // Email addresses
+  [
+    /(^|[ \t\r\n\v\f])([a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]{1,64}@[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(\.[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?){1,500})/gim,
+    '$1<email>',
+  ],
 
   // Generic user directories
   [/\/(Users?)\/[^/]*\//gm, '/$1/<user>/'],
```
addaleax, line 31: This can't be reached anymore, can it?

addaleax (approved):

PR #347, opened 2024-06-06
feat!(devtools-connect): update oidc MONGOSH-1790

mabaasit:
## Checklist
- [ ] I have signed the Contributor License Agreement (https://www.mongodb.com/legal/contributor-agreement)
- [ ] New tests and/or benchmarks are included
- [ ] Documentation is changed or added

PR #407, opened 2024-07-11
fix(devtools-connect): bump `mongodb` to latest and fix OIDC option check MONGOSH-1831

addaleax:
Missed this option update in https://github.com/mongodb-js/devtools-shared/pull/347 because we didn't bump the driver (and therefore also not its types) in it

### DIFF
--- a/packages/devtools-connect/src/connect.spec.ts
+++ b/packages/devtools-connect/src/connect.spec.ts
@@ -429,20 +429,20 @@ describe('devtools connect', function () {
       ).to.equal(true);
     });
 
-    it('returns false if the PROVIDER_NAME JS option is set', function () {
+    it('returns false if the ENVIRONMENT JS option is set', function () {
       expect(
         isHumanOidcFlow('mongodb://example/?authMechanism=MONGODB-OIDC', {
           authMechanismProperties: {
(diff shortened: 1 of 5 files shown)


packages/devtools-connect/src/connect.spec.ts
```diff
@@ -429,20 +429,20 @@ describe('devtools connect', function () {
       ).to.equal(true);
     });
 
-    it('returns false if the PROVIDER_NAME JS option is set', function () {
+    it('returns false if the ENVIRONMENT JS option is set', function () {
       expect(
         isHumanOidcFlow('mongodb://example/?authMechanism=MONGODB-OIDC', {
           authMechanismProperties: {
-            PROVIDER_NAME: 'aws',
+            ENVIRONMENT: 'azure',
```
kmruiz, line 436: Why did we change the value from aws to azure?

kmruiz (approved):

packages/devtools-connect/src/connect.spec.ts
addaleax: Because `aws` is not actually a valid value for this 🙂 https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#mongocredential-properties-6

PR #418, opened 2024-07-18
feat(devtools-proxy-support): initial shared implementation COMPASS-8072

addaleax:
A few things small that might be helpful for reviewing this PR:

Some potentially helpful reading: 
- [Technical design](https://docs.google.com/document/d/1smhTMlNIotueMJPcEsBEG0q1BmwvxQIpEcGIeLj8bwY/edit)
- MDN entry for [`CONNECT` proxies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT)
- Wikipedia entry for [`PAC (proxy auto-config) files`](https://en.wikipedia.org/wiki/Proxy_auto-config)
- [Node.js HTTP Agent documentation](https://nodejs.org/api/http.html#class-httpagent) (tl;dr: An `Agent` is a connection/socket factory)
- [`proxy-agent` package README](https://github.com/TooTallNate/proxy-agents/tree/main/packages/proxy-agent)

This PR will replace:
- [`@mongosh/import-node-fetch`](https://github.com/mongodb-js/mongosh/blob/8e6962432397154941f593c847d8f774bfd49f1c/packages/import-node-fetch/src/index.ts) in mongosh
- [`@mongodb-js/ssh-tunnel`](https://github.com/mongodb-js/compass/tree/55a5a608713d7316d158dc66febeb6b114d8b40d/packages/ssh-tunnel/src) in Compass (split into two things, `SSHAgent` and `Socks5Server`)

I'd recommend starting with `proxy-options.ts` and `agent.ts` and their associated test files, then work onwards from there.


packages/devtools-proxy-support/src/agent.ts
```diff
@@ -0,0 +1,116 @@
...
+  target?: string
+): AgentWithInitialize | undefined {
+  if ('createConnection' in proxyOptions) {
+    return proxyOptions as AgentWithInitialize;
+  } else {
+    if (
+      target !== undefined &&
+      !proxyForUrl(proxyOptions as DevtoolsProxyOptions, target)
+    )
+      return undefined;
```
gribnoysup, line 113: Super nit: hard to read these sequential returns otherwise

```suggestion
    if (
      target !== undefined &&
      !proxyForUrl(proxyOptions as DevtoolsProxyOptions, target)
    ) {
      return undefined;
    }
```

packages/devtools-proxy-support/src/agent.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  override async connect(
+    req: ClientRequest,
+    opts: AgentConnectOpts
+  ): Promise<HTTPAgent> {
+    if (this.sshAgent) return this.sshAgent;
+    while (this._reqLock) {
+      await this._reqLock;
+    }
```
gribnoysup, line 84: Just curious why we are using while here: is there a possibility of a microtask happening between await resolving and the execution going to the beginning of the loop to check the `_reqLock`?

packages/devtools-proxy-support/src/socks5.ts
```diff
@@ -0,0 +1,382 @@
...
+        {
+          host: dstAddr,
+          port: dstPort,
+        },
+        (err, sock) => {
+          // Ideally, we would always be using this callback for retrieving the `sock`
+          // instance. However, agent-base does not call the callback at all if
+          // the agent resolved to another agent (as is the case for e.g. `ProxyAgent`).
+          if (err) reject(err);
+          else if (sock) resolve(sock);
```
alenakhineika, line 260: What if there is no error but sock is undefined? this promise remains unresolved?

addaleax: If neither this line nor L247 are triggered with a valid socket, then yeah, the Promise would remain unresolved.

Tbh, I think I did this to appease typescript – either `err` or `sock` should always be truth-y here

addaleax: c6d99f8654f8ee3bc407272c91c62c47633b69b4 should help with this?

gribnoysup (approved):
I left some small nits and one question just for my understanding, otherwise it looks good, but it's a lot of code, so I definitely might've missed some things 😅

packages/devtools-connect/src/connect.ts
```diff
@@ -386,82 +419,163 @@ export async function connectMongoClient(
...
-    (client as any).autoEncrypter[
-      Symbol.for('@@mdb.decorateDecryptionResult')
-    ] = true;
+}
+
+function hasProxyHostOption(
+  uri: string,
+  clientOptions: MongoClientOptions
+): boolean {
+  if (clientOptions.proxyHost || clientOptions.proxyPort) return true;
```
alenakhineika, line 569: Maybe we could rename this function to `hasProxyHostOrPortOption` option? In which cases could it be that proxyPort is enough without a host? Will be a parent host used then?

packages/devtools-proxy-support/src/fetch.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { RequestInit, RequestInfo, Request, Response } from 'node-fetch';
+import type { AgentWithInitialize } from './agent';
+import { useOrCreateAgent } from './agent';
+import type { DevtoolsProxyOptions } from './proxy-options';
+
+declare const __webpack_require__: unknown;
+
+// The original version of this code was largely taken from
+// https://github.com/mongodb-js/mongosh/blob/8e6962432397154941f593c847d8f774bfd49f1c/packages/import-node-fetch/src/index.ts
+async function importNodeFetch(): Promise<typeof fetch> {
```
alenakhineika, line 11: 🔥

packages/devtools-proxy-support/src/agent.ts
addaleax: > is there a possibility of a microtask happening between await resolving and the execution going to the beginning of the loop to check the `_reqLock`?

The `await` here does always wait at least one microtask queue iteration, and during that iteration other things can happen (such as other Promises being resolved and handlers for that resolution running)

> Just curious why we are using while here

Yeah, it's to avoid a scenario where if `.connect()` gets called twice (or more) in a synchronous code block, both of those `.connect()` calls would await the same `_reqLock` Promise and then simultaneously proceed (and each of them create new, independent `_reqLock` promises, overwriting each other)

packages/devtools-connect/src/connect.ts
addaleax: `proxyHost` is the authoritative option, the driver forbids using `proxyPort` without `proxyHost`, e.g.

```sh
$ mongosh --quiet 'mongodb://localhost/?proxyPort=1234'
MongoParseError: Must specify proxyHost if other proxy options are passed
```

We could drop the `proxyPort` check here, I felt like it made sense to have it in case the driver ever changes this behavior (and defaults to e.g. `localhost`), but not feeling strongly about that at all

packages/devtools-proxy-support/src/ssh.ts
```diff
@@ -0,0 +1,177 @@
...
+      keepaliveInterval: 20000,
+      host: this.url.hostname,
+      port: +this.url.port || 22,
+      username: this.url.username || undefined,
+      password: this.url.password || undefined,
+      privateKey: this.proxyOptions.sshOptions?.identityKeyFile
+        ? await fs.readFile(this.proxyOptions.sshOptions.identityKeyFile)
+        : undefined,
+      passphrase: this.proxyOptions.sshOptions?.identityKeyPassphrase,
+      // debug: console.log.bind(null, '[client]')
```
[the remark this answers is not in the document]
gribnoysup: Gothca, makes sense 😄

alenakhineika (approved):
Wow Anna, this is amazing! I rely more on your tests than my eyes 😄 but did my best reading through your changes a couple of times and don't see anything else except these comments I have already left. Great work! 😍

packages/devtools-proxy-support/src/agent.ts
gribnoysup: Aha, got it, thanks for a detailed explanation with an example, it like made sense, but was still hard to wrap my head around the details of the flow

addaleax: Yeah ... it didn't feel worth pulling in an external library for this, but if you feel like that would be a good idea to make it more obvious what's happening here, I am not opposed to that at all (because it's not just you, I also had to give this a bit of thought to be confident that it's correct the way it is)

addaleax: Sure, added a small comment about this as well 👍

PR #427, opened 2024-07-29
chore(oidc-mock-provider): allow binding to all ips while keeping a specific hostname for the issuer

lerouxb:
`--bind_ip_all` and `bindIpAll` chosen to match the [server's config](https://www.mongodb.com/docs/manual/core/security-mongodb-configuration/).

This helps with docker host networking so that the mongodb server in docker can get to the oidc server you run on the host or another docker container while the issuer remains just localhost.

### DIFF
--- a/packages/oidc-mock-provider/src/index.ts
+++ b/packages/oidc-mock-provider/src/index.ts
@@ -61,6 +61,11 @@ export interface OIDCMockProviderConfig {
   hostname?: string;
 
   /**
+   * Optional bind to all IPv4 and IPv6 addresses.
+   */
+  bindIpAll?: boolean;
+
+  /**
    * Optional additional fields to be returned when the OIDC configuration is accessed.
    */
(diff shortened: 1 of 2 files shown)


packages/oidc-mock-provider/src/index.ts
```diff
@@ -102,7 +107,10 @@ export class OIDCMockProvider {
   }
 
   private async init(): Promise<this> {
-    this.httpServer.listen(this.config.port ?? 0, this.config.hostname);
+    this.httpServer.listen(
+      this.config.port ?? 0,
+      this.config.bindIpAll ? '::,0.0.0.0' : this.config.hostname
```
addaleax, line 112: > `Uncaught Error: getaddrinfo ENOTFOUND ::,0.0.0.0`

This should probably be/can just be

```suggestion
      this.config.bindIpAll ? '::' : this.config.hostname
```

addaleax (approved):

PR #439, opened 2024-08-03
build: add TypeScript to CodeQL

rtritto:
## Description

Add TypeScript language to CodeQL.

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -33,7 +33,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'javascript' ]
+        language: [ 'javascript-typescript' ]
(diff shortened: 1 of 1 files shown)


addaleax:
@rtritto If I understand the [GH docs for CodeQL](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#changing-the-languages-that-are-analyzed) correctly, these are just aliases for each other – just want to check, is the intention of this PR to be explicit about the fact that both JS and TS are being analyzed (and, in that case, would we prefer `javascript-typescript` as a single value to make it more obvious that that's what's happening, and not two separate analyses)? Or did something actually change w.r.t. the meaning of these labels and are the GH docs out of date?

rtritto:
Currently projects are JS and TS, so we need both in CodeQL.
Can be used a single label (`javascript-typescript`), I done the commit.

PR #445, opened 2024-08-07
feat(node-webpack-startup-snapshot-checker): add package

addaleax:
Add a script that tells you whether a module is simple enough or not to be included in a Node.js startup snapshot.

### DIFF
--- a/packages/node-webpack-startup-snapshot-checker/src/index.ts
+++ b/packages/node-webpack-startup-snapshot-checker/src/index.ts
@@ -1,0 +1,133 @@
+/* eslint-disable no-console */
+import { execFile as execFileCallback } from 'child_process';
+import { promisify } from 'util';
+import os from 'os';
+import path from 'path';
+import { promises as fs } from 'fs';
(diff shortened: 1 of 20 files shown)


packages/node-webpack-startup-snapshot-checker/src/index.ts
```diff
@@ -0,0 +1,133 @@
...
+  output: {
+    path: path.resolve(__dirname, 'dist'),
+    filename: 'webpacked.js',
+    library: { name: 'entry', type: 'var' }
+  },
+  module: {
+    rules: [
+      {
+        test: /\\.node$/,
+        loader: "null-loader",
```
alenakhineika, line 70: Do we want to silence all node modules? Or what exactly here is happening?

addaleax: Yup, exactly that 👍 They can't be snapshotted anyway

alenakhineika (approved):

PR #446, opened 2024-08-07
fix(devtools-proxy-support): lazily load expensive dependencies MONGOSH-1853

addaleax:
Notably, this is referring to `pac-proxy-agent` and `ssh2`, which both involve somewhat complex setups.

Also, since `ssh2` includes a WebAssembly implementation of a cryptographic algorithm, we should make sure not to load it in FIPS mode.

(This does not fully remove the need for loading this package lazily, since it still relies on the `http` package directly.)

## Checklist

### DIFF
--- a/packages/devtools-proxy-support/src/ssh.ts
+++ b/packages/devtools-proxy-support/src/ssh.ts
@@ -18,7 +18,7 @@ function ssh2(): typeof import('ssh2') {
   if (getFips()) {
     // ssh2 uses a WASM implementation of the non-FIPS-compliant Poly1305 hash algorithm
     throw new Error(
-      'Using `ssh2` features in FIPS mode is currently not available'
+      'devtools-proxy-support: Using `ssh2` features in FIPS mode is currently not available'
     );
   }
   // Lazily loading this package because it uses WebAssembly and therefore cannot


packages/devtools-proxy-support/src/ssh.ts
```diff
@@ -4,14 +4,28 @@ import type { DevtoolsProxyOptions } from './proxy-options';
...
 import type { ProxyLogEmitter } from './logging';
 import { connect as tlsConnect } from 'tls';
 import type { Socket } from 'net';
+import { getFips } from 'crypto';
+
+// eslint-disable-next-line @typescript-eslint/consistent-type-imports
+function ssh2(): typeof import('ssh2') {
+  if (getFips()) {
+    // ssh2 uses a WASM implementation of the non-FIPS-compliant Poly1305 hash algorithm
+    throw new Error(
```
kmruiz, line 20: I might be wrong, but I feel this is the kind of error that we will find somewhere in a JIRA ticket and we will need to add logs later to understand a bit more. Is there any useful information we can add to the exception message?

packages/devtools-proxy-support/src/proxy-agent.ts
```diff
@@ -0,0 +1,210 @@
...
+  socks5h: [wellKnownAgents.socks, wellKnownAgents.socks],
+  'pac+data': [wellKnownAgents.pac, wellKnownAgents.pac],
+  'pac+file': [wellKnownAgents.pac, wellKnownAgents.pac],
+  'pac+ftp': [wellKnownAgents.pac, wellKnownAgents.pac],
+  'pac+http': [wellKnownAgents.pac, wellKnownAgents.pac],
+  'pac+https': [wellKnownAgents.pac, wellKnownAgents.pac],
+};
+
+function isValidProtocol(v: string): v is ValidProtocol {
+  return Object.keys(proxies).includes(v);
```
kmruiz, line 98: Why can't we just do `!!proxies[v]`? I know it's not very important in this case, but just to make sure I understand this part of the code.

kmruiz (approved):

packages/devtools-proxy-support/src/proxy-agent.ts
addaleax: Oh, literally just kept this as it's a bit closer to the original code. No strong preferences tbh

packages/devtools-proxy-support/src/ssh.ts
addaleax: Hm yeah, I've had the same thought but haven't really come to a conclusion yet. I've added the package name here in 388c27a3851d5ac733a3e4099572a9c262acd25e, that makes it hopefully a bit easier to track down if this ever comes up.

Currently, I'm not too worried, because:
- SSH support is only explicitly exposed in Compass
- FIPS support is only explicitly exposed in mongosh

so at least in the current world, we shouldn't expect too much trouble coming out of this.

packages/devtools-proxy-support/src/proxy-agent.ts
kmruiz: Ah then nothing, I thought there was some rewrite. Then let's keep it as the original.

alenakhineika (approved):

packages/devtools-proxy-support/src/ssh.ts
kmruiz: That sounds good enough for me! Is there any situation (don't know much how FIPS work) where a user could unknowingly activate FIPS support or an option to tell them why FIPS is activated? If not (to any of the questions haha) just feel free to resolve this comment and merge.

addaleax: For mongosh, they'd typically either specify `--tlsFIPSMode` explicitly or call `crypto.setFips(1)` themselves.

I _guess_ there's a chance that this can be configured through the system OpenSSL config, but honestly, I don't think we've ever had any issues of that kind reported to us

PR #467, opened 2024-09-05
fix(devtools-proxy-support): remove certificates without issuer from system CA list COMPASS-8252

addaleax:
We have discovered that system certificate stores may contain certificates that a) OpenSSL may choose over other roots for the same certificate chain and b) are not accompanied by their own root certificate, i.e. only present an immediate certificate.

While we will ideally solve this by using `X509_V_FLAG_PARTIAL_CHAIN`, Node.js does not expose that as a feature yet.

## Checklist


packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -58,12 +107,14 @@ export async function systemCA(
   existingOptions: {
     ca?: NodeJSCAOption;
     tlsCAFile?: string | null | undefined;
-  } = {}
+  } = {},
+  allowCertificatesWithoutIssuer?: boolean // defaults to false
```
gribnoysup, line 111: Do we want to also expose this in connect method options or is the idea that we'd allow to activate it with the env var?

gribnoysup (approved):

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -50,6 +51,54 @@ export function mergeCA(...args: (NodeJSCAOption | undefined)[]): string {
...
+      let parsed: X509Certificate | null = null;
+      try {
+        parsed = new X509Certificate(pem);
+      } catch (err: unknown) {
+        messages.push(
+          `Unable to parse certificate: ${
+            err && typeof err === 'object' && 'message' in err
+              ? String(err.message)
+              : String(err)
+          }`
```
nirinchev, line 79: No strong opinion here and mostly out of curiosity, can this be rewritten more concisely as: `Unable to parse certificate: ${ err?.message || err }`? I guess it will require replacing the type for `err` with `any`, but other than that are there downsides?

packages/devtools-proxy-support/src/system-ca.ts
addaleax: If we wanted to expose it the idea is that we'd forward it all the way to the user in the UI, right?

I am pretty hopeful that this won't be necessary, and it's a temporary solution anyway. So yeah, the environment variable would be the way to get back to the current (i.e. "broken" 1.44) behavior if somebody does end up needing that for some reason

gribnoysup: Ohhhhh, wait, I totally misread the variable name and thought it's the other way around, okay yeah, not exposing it at all makes sense 😄

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -50,6 +51,54 @@ export function mergeCA(...args: (NodeJSCAOption | undefined)[]): string {
...
+        );
+      }
+      return { pem, parsed };
+    });
+  pemWithParsedCache.set(ca, caWithParsedCerts);
+  caWithParsedCerts = caWithParsedCerts.filter(({ parsed }) => {
+    const keep =
+      !parsed ||
```
nirinchev, line 87: This is somewhat counterintuitive to me - if we were unable to parse the certificate, do we want to keep it? I guess there's no harm in doing it, but do we feel it's useful to hold on to pems that are not valid x509 certificates?

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -50,6 +51,54 @@ export function mergeCA(...args: (NodeJSCAOption | undefined)[]): string {
...
+          `Unable to parse certificate: ${
+            err && typeof err === 'object' && 'message' in err
+              ? String(err.message)
+              : String(err)
+          }`
+        );
+      }
+      return { pem, parsed };
+    });
+  pemWithParsedCache.set(ca, caWithParsedCerts);
```
nirinchev, line 84: Likely inconsequential, but for my understanding, why are we storing the unfiltered certificates in the cache rather than the filtered ones? My expectation is that the result of the filter will always be the same, so we might as well cache that rather than do the filtering every time.

lerouxb (approved):
Nice! Really quick turnaround time 😄

nirinchev (commented):
This change looks reasonable to me, though I only have superficial context via the HELP tickets and the slack comms, so worth getting another pair of eyes on it.

packages/devtools-proxy-support/src/system-ca.ts
addaleax: > I guess it will require replacing the type for `err` with `any`, but other than that are there downsides?

Nah, you pretty much got it – it's not more than a general tendency to avoid `any` whenever we can 🙂

packages/devtools-proxy-support/src/system-ca.ts
addaleax: I don't think we should ever encounter a situation like this in the first place. But if we do, I think this is the "safer" path – if nothing else, it aligns more closely with the previous behavior (and the one we're aiming for in the long term, once this code is removed again). So I think just leaving a diagnostic message here is probably okay?

packages/devtools-proxy-support/src/system-ca.ts
addaleax: Good idea! Done in 4b60c714f082f69355a7a5ed66f15980b17362f9

PR #469, opened 2024-09-05
fix(devtools-proxy-support): handle special characters in SSH URL correctly COMPASS-8254

addaleax:
This is unfortunately very obvious in retrospect 😞

### DIFF
--- a/packages/devtools-proxy-support/src/ssh.ts
+++ b/packages/devtools-proxy-support/src/ssh.ts
@@ -78,10 +78,10 @@ export class SSHAgent extends AgentBase implements AgentWithInitialize {
     const sshConnectConfig: ConnectConfig = {
       readyTimeout: 20000,
       keepaliveInterval: 20000,
-      host: this.url.hostname,
+      host: decodeURIComponent(this.url.hostname),
       port: +this.url.port || 22,
-      username: this.url.username || undefined,
-      password: this.url.password || undefined,
+      username: decodeURIComponent(this.url.username) || undefined,
+      password: decodeURIComponent(this.url.password) || undefined,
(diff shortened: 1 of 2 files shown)


Anemy (approved):

packages/devtools-proxy-support/src/ssh.ts
```diff
@@ -92,7 +92,7 @@ export class SSHAgent extends AgentBase implements AgentWithInitialize {
     this.logger.emit('ssh:establishing-conection', {
       host: sshConnectConfig.host,
       port: sshConnectConfig.port,
-      password: !!sshConnectConfig.passphrase,
+      password: !!sshConnectConfig.password,
```
lerouxb, line 95: 😭

lerouxb (approved):

packages/devtools-proxy-support/src/ssh.ts
addaleax: To be fair, this bit is kinda unrelated, but still, yeah, not my smartest move 😅

lerouxb: Yeah I see now it is just in a logging event 👍

packages/devtools-proxy-support/src/ssh.ts
```diff
@@ -78,10 +78,10 @@ export class SSHAgent extends AgentBase implements AgentWithInitialize {
     const sshConnectConfig: ConnectConfig = {
       readyTimeout: 20000,
       keepaliveInterval: 20000,
-      host: this.url.hostname,
+      host: decodeURIComponent(this.url.hostname),
```
lerouxb, line 81: I can see how this would be easy to miss.

PR #474, opened 2024-09-20
chore(devtools-proxy-support): push expired certs to the bottom of the system CA list COMPASS-8322

gribnoysup:
We ran into more weird issues with CA cert validation failing in unexpected cases (see COMPASS-8322 for more details) affecting users because system CAs are now added to the CA list by default. To work around this particular one we're sorting a list of CAs returned from system store by expiration date to make sure that the ones that will expire sooner or expired already have less of a chance to get picked up from the CA list

### DIFF
--- a/packages/devtools-proxy-support/src/system-ca.ts
+++ b/packages/devtools-proxy-support/src/system-ca.ts
@@ -68,6 +68,10 @@ export function parseCACerts(
     try {
       parsed = new X509Certificate(pem);
     } catch (err: unknown) {
+      // Most definitely should happen never or extremely rarely, in case it
+      // does, if this cert will affect the TLS connection verification, the
+      // connection will most definitely fail and we'll se it in the logs. For
+      // that reason we're just logging, but not throwing an error here
       messages.push(
         `Unable to parse certificate: ${
           err && typeof err === 'object' && 'message' in err
@@ -80,15 +84,14 @@ export function parseCACerts(
   });
 }
 
-function doesCertificateHasMatchingIssuer(
-  { parsed }: ParsedX509Cert,
+function certificateHasMatchingIssuer(
+  cert: X509Certificate,
   certs: ParsedX509Cert[]
 ) {
   return (
-    !parsed ||
-    parsed.checkIssued(parsed) ||
+    cert.checkIssued(cert) ||
     certs.some(({ parsed: issuer }) => {
-      return issuer && parsed.checkIssued(issuer);
+      return issuer && cert.checkIssued(issuer);
     })
   );
 }
@@ -123,7 +126,11 @@ export function removeCertificatesWithoutIssuer(
 
   const _messages: string[] = [];
   const filteredCAlist = ca.filter((cert) => {
-    const keep = doesCertificateHasMatchingIssuer(cert, ca);
+    // If cert was not parsed, we want to keep it in the list. The case should
+    // be generally very rare, but in case it happens and this cert will affect
+    // the TLS handshake, it will show up in the logs as the connection error
+    // anyway, so it's safe to keep it
+    const keep = !cert.parsed || certificateHasMatchingIssuer(cert.parsed, ca);
     if (!keep && cert.parsed) {
       const { parsed } = cert;
       _messages.push(
--- a/packages/devtools-proxy-support/src/system-ca.ts
+++ b/packages/devtools-proxy-support/src/system-ca.ts
@@ -70,7 +70,7 @@ export function parseCACerts(
     } catch (err: unknown) {
       // Most definitely should happen never or extremely rarely, in case it
       // does, if this cert will affect the TLS connection verification, the
-      // connection will most definitely fail and we'll se it in the logs. For
+      // connection will most definitely fail and we'll see it in the logs. For
       // that reason we're just logging, but not throwing an error here
       messages.push(
         `Unable to parse certificate: ${
--- a/packages/devtools-proxy-support/src/system-ca.ts
+++ b/packages/devtools-proxy-support/src/system-ca.ts
@@ -31,50 +31,39 @@ function systemCertsCached(systemCAOpts: SystemCAOptions = {}): Promise<{
   return systemCertsCachePromise;
 }
 
+function certToString(cert: string | Uint8Array) {
+  return typeof cert === 'string'
+    ? cert
+    : Buffer.from(cert.buffer, cert.byteOffset, cert.byteLength).toString(
+        'utf8'
+      );
+}
+
 export function mergeCA(...args: (NodeJSCAOption | undefined)[]): string {
   const ca = new Set<string>();
   for (const item of args) {
     if (!item) continue;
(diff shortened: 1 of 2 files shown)


packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -150,14 +216,14 @@ export async function systemCA(
...
-    const reducedList = removeCertificatesWithoutIssuer(systemCerts);
-    systemCerts = reducedList.ca;
-    messages = messages.concat(reducedList.messages);
+    systemCerts = removeCertificatesWithoutIssuer(systemCerts, messages);
   }
 
   return {
     ca: mergeCA(
-      systemCerts,
+      sortByExpirationDate(systemCerts).map((cert) => {
```
gribnoysup, line 224: One thing I considered but haven't pursued is sorting all the certs we're passing here, I don't think it will have a major effect, but it seems like it would be closer to what's happening when openssl gets a list of certs passed. I had to restructure the code and split the functions in a way that should allow to do it pretty easy if we decide to change that.

Also not exactly sure about the ordering of passed arguments here, taking into account that it seems like it's important in what order certs are being passed to TLS connection, it might be that we should reverse the order and give priority to whatever user is passing directly through connection form, then certs bundled with Node.js, and only then the system certs

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -31,50 +31,39 @@ function systemCertsCached(systemCAOpts: SystemCAOptions = {}): Promise<{
...
-              'utf8'
-            );
-      ca.add(asString);
+      ca.add(certToString(cert));
     }
   }
   return [...ca].join('\n');
 }
 
-const pemWithParsedCache = new WeakMap<
```
gribnoysup, line 54: This cache is still here, just got moved down with the function and renamed to clarify the purpose

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -31,50 +31,39 @@ function systemCertsCached(systemCAOpts: SystemCAOptions = {}): Promise<{
...
-    | undefined = pemWithParsedCache.get(ca);
+export type ParsedX509Cert = { pem: string; parsed: X509Certificate | null };
 
-  const messages: string[] = [];
-  let caWithParsedCerts = ca.map((pem) => {
+/**
+ * Safely parse provided certs, push any encountered errors to the provided
+ * messages array
+ */
+export function parseCACerts(
```
gribnoysup, line 60: Split parsing out of removing function so that I can parse once, then optionally remove, then use the same parsed certs when sorting without loosing the parsed certs inside the "removing" function

gribnoysup:
Marking as draft because I was looking at the wrong cert all this time when trying to confirm that the fix worked and there seems to be more to it: if I sort all the certs by validTo, connection succeeds, but if I just push the "expired" ones to the bottom, it doesn't, so trying to very carefully confirm again why exactly sorting helps with fixing this

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -150,14 +203,14 @@ export async function systemCA(
...
     )
   ) {
-    const reducedList = removeCertificatesWithoutIssuer(systemCerts);
-    systemCerts = reducedList.ca;
-    messages = messages.concat(reducedList.messages);
+    systemCerts = removeCertificatesWithoutIssuer(systemCerts, messages);
   }
 
   return {
     ca: mergeCA(
```
gribnoysup, line 210: Not something I'm planning to pursue right now, but assuming openssl doesn't fail when using the same CA list that we used to repro the issue because they are doing the sorting somewhere internally and Node.js doesn't, it seems like we might want to sort all the options combined, and not only the ones coming from system CA, but I'm not feeling confident enough to do it, would like to confirm exactly why this is not happening when trying the same connection with openssl directly and not with Node.js tls module

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -89,23 +78,84 @@ export function removeCertificatesWithoutIssuer(ca: string[]): {
...
-      !parsed ||
-      parsed.checkIssued(parsed) ||
-      caWithParsedCerts.find(
-        ({ parsed: issuer }) => issuer && parsed.checkIssued(issuer)
-      );
-    if (!keep) {
-      messages.push(
+}
+
+function doesCertificateHasMatchingIssuer(
```
lerouxb, line 83: grammar nit: have, not has :)

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -89,23 +78,84 @@ export function removeCertificatesWithoutIssuer(ca: string[]): {
...
+  { parsed }: ParsedX509Cert,
+  certs: ParsedX509Cert[]
+) {
+  return (
+    !parsed ||
```
lerouxb, line 88: So if parsed is falsey we return that it does have a matching issuer?

lerouxb: I suppose that's old code.

lerouxb (approved):

packages/devtools-proxy-support/src/system-ca.ts
gribnoysup: Yes, I'm just keeping it as-is, but FWIW [Annas reasoning](https://github.com/mongodb-js/devtools-shared/pull/467#discussion_r1745678311) in the PR that introduced it seems fair to me. I'll add it as a comment here so that we don't need to track down the PR every time

packages/devtools-proxy-support/src/system-ca.ts
gribnoysup: TY! Will fix

lerouxb: so either doesCertificateHaveMatchingIssuer or certificateHasMatchingIssuer

packages/devtools-proxy-support/src/system-ca.ts
gribnoysup: Ended up moving the `!parsed` check outside of the method just so this function name clearly describes and matches what it does. Moved the not parsed check to the filter itself and added a comment, seems a bit cleaner that way?

packages/devtools-proxy-support/src/system-ca.ts
```diff
@@ -89,23 +82,87 @@ export function removeCertificatesWithoutIssuer(ca: string[]): {
...
+    return result.ca;
+  }
+
+  const _messages: string[] = [];
+  const filteredCAlist = ca.filter((cert) => {
+    // If cert was not parsed, we want to keep it in the list. The case should
+    // be generally very rare, but in case it happens and this cert will affect
+    // the TLS handshake, it will show up in the logs as the connection error
+    // anyway, so it's safe to keep it
+    const keep = !cert.parsed || certificateHasMatchingIssuer(cert.parsed, ca);
```
lerouxb, line 133: ❤️

PR #478, opened 2024-09-29
fix(shell-bson-parser): Support `valueOf` date method

benleembruggen:
## Description

Adds support for the [`valueOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf) Date method to the `shell-bson-parser` package

## Checklist

### DIFF
--- a/packages/shell-bson-parser/src/index.spec.ts
+++ b/packages/shell-bson-parser/src/index.spec.ts
@@ -425,6 +425,7 @@ describe('@mongodb-js/shell-bson-parser', function () {
           setUTCSeconds: (${newDate}).setUTCSeconds(59),
           setYear: (${newDate}).setYear(96),
           toISOString: (${newDate}).toISOString(),
+          valueOf: (${newDate}.valueOf()),
        }`;
               expect(parse(input, options)).to.deep.equal({
                 getDate: new (Date as any)(...args).getDate(),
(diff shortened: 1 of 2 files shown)


mcasimir (approved):

mcasimir:
Nit: worth adding a test?

mongo-J:
Hi @mcasimir, do you know by any chance that what is the process of releasing the patch for the package or is it fully automated once pr is merged. And if there is any requirement for documentation? Also the PR checklist has a contributor license agreement does @benleembruggen need to sign this? TIA!

Anemy:
@mongo-J 
> what is the process of releasing the patch for the package
It's fully automated when a pr is merged. We should probably add a bit of documentation for that. I'll open a pr. The versioning (patch/minor/major) happens depending on the commit message. It uses conventional commits: https://www.conventionalcommits.org/en/v1.0.0/  This pr looks fine as is though, it will be a minor bump when merged.

> And if there is any requirement for documentation
We don't keep changelogs for these packages, so no. If it's a breaking change or needs a migration then it can be good to add that to a README somewhere or add a changelog describing it. In that case we also would want a major version bump, which can be indicated through a `!`. https://www.conventionalcommits.org/en/v1.0.0/

>  Also the PR checklist has a contributor license agreement does @benleembruggen need to sign this?
If they are a MongoDB employee no.


The code behind bumping packages if curious: https://github.com/mongodb-js/devtools-shared/blob/main/packages/monorepo-tools/src/bump-packages.ts

PR #488, opened 2024-11-11
fix: Suppress node 23 warnings about require-ing ESM MONGOSH-1895

nirinchev:
## Description

Starting with node 23, support for require-ing ES modules no longer throws but instead prints a warning in the output. This is problematic for homebrew users since homebrew aggressively upgrades node, regardless of what the preferred node version is for a package. This PR takes one approach, which is to suppress experimental warnings altogether for Node 23. This is only manually tested since I'm not convinced it's the approach we'd like to take (see Open Questions) - once we've settled on it, I'll need to figure out a way to test. 

## Open Questions
This feels a bit hacky and I'm not convinced this is the long term fix we'd like to use. Here are a few alternatives, ranked from easy to hard:
1. We could try and more narrowly suppress the specific require() warning.
2. We could restore the original warning emitter after importing fetch.
3. We could change `moduleResolution` to `NodeNext` and fix the issues that this causes (unfortunately, I wan't able to trace what they were).
4. We could migrate to ES modules.

## Checklist

### DIFF
--- a/packages/devtools-proxy-support/src/fetch.ts
+++ b/packages/devtools-proxy-support/src/fetch.ts
@@ -12,11 +12,12 @@ async function importNodeFetch(): Promise<typeof fetch> {
   // Node-fetch is an ESM module from 3.x
   // Importing ESM modules to CommonJS is possible with a dynamic import.
   // However, once this is transpiled with TS, `await import()` changes to `require()`, which fails to load
-  // the package at runtime.
+  // the package at runtime for Node < 23.
   // The alternative, to transpile with "moduleResolution": "NodeNext", is not always feasible.
   // Use this function to safely import the node-fetch package
   let module: typeof fetch | { default: typeof fetch };
   try {
+    suppressExperimentalWarningsIfNecessary();
     module = await import('node-fetch');
   } catch (err: unknown) {
     if (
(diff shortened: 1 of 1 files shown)


packages/devtools-proxy-support/src/fetch.ts
```diff
@@ -37,6 +38,35 @@ async function importNodeFetch(): Promise<typeof fetch> {
...
+      ) {
+        return;
+      }
+
+      originalEmit(warning, ...args);
+    };
+
+    warningsSuppressed = true;
+  }
+}
```
addaleax, line 68: If this is the approach we're going with – which I think is pretty reasonable – then I'd do that in the mongosh CLI itself, like we already disable warnings for the case of the compiled executable in `packages/cli-repl/src/run.ts`, because that's where we're expecting overrides for Node.js behavior and other modifications to truly global state, how does that sound to you?

nirinchev: Sure, that sounds reasonable to me! I guess originally I wanted to put it next to the code that does the requiring shenanigans, but you're right that the warnings being displayed or not is better left to the user-facing application.

addaleax: Yeah, I think there's also just going to be _more_ code running into this problem in the future, the more popular ESM gets in our dependencies, not just the code here

nirinchev:
Closing in favor of https://github.com/mongodb-js/mongosh/pull/2258.

PR #492, opened 2024-11-27
chore(ci): use github app instead of bot user

nirinchev:
## Description
This refreshes our ci workflows and switches to using github app tokens instead of the bot user we used to use.

### DIFF
--- a/.github/workflows/bump-packages.yaml
+++ b/.github/workflows/bump-packages.yaml
@@ -10,25 +10,39 @@ jobs:
     name: Bump packages
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.DEVTOOLS_BOT_APP_ID }}
+          private-key: ${{ secrets.DEVTOOLS_BOT_PRIVATE_KEY }}
+
+      - name: Get GitHub App User ID
+        id: get-user-id
+        run: echo "user-id=$(gh api "/users/${{ steps.app-token.outputs.app-slug }}[bot]" --jq .id)" >> "$GITHUB_OUTPUT"
+        shell: bash
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+
+      - name: Set up Git
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/bump-packages.yaml
```diff
@@ -10,25 +10,39 @@ jobs:
     name: Bump packages
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/create-github-app-token@v1
```
nirinchev: Here we can't use the [setup-bot-token](https://github.com/mongodb-js/devtools-shared/blob/main/actions/setup-bot-token/action.yml) action because it's in the same repo and we must checkout it first to be able to use it.

.github/workflows/bump-packages.yaml
```diff
@@ -46,21 +60,4 @@ jobs:
...
-            - Bump package versions
-
-      - name: Merge PR
-        if: ${{steps.cpr.outputs.pull-request-operation == 'created'}}
-        env:
-          PULL_REQUEST_NUMBER: ${{steps.cpr.outputs.pull-request-number}}
-          GITHUB_TOKEN: ${{secrets.PAT}}
-        run: |
-          gh pr merge $PULL_REQUEST_NUMBER --squash --delete-branch
+          git push
```
nirinchev, line 51: Pushing directly to main instead of creating a PR to align with our other workflows.

.github/workflows/update-cidrs.yaml
```diff
@@ -11,42 +11,35 @@ jobs:
...
-          npm install -g npm@8
+      - uses: actions/checkout@v4
+        with:
+          # don't checkout a detatched HEAD
+          ref: ${{ github.head_ref }}
+          token: ${{ steps.app-token.outputs.token }}
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ^20.x
```
nirinchev, line 28: The action has been failing for a while because node 18 is not a supported engine.

nirinchev (commented):
There are some whitespace changes due to prettier reformatting the yaml, so this is best reviewed with the "hide whitespace" option.

PR #498, opened 2025-01-17
feat(mongodb-log-writer): move mongodb-log-writer into devtools-shared MONGOSH-1991

gagik:
This also includes slight refactoring to split the modules of MongoLogManager and MongoLogWriter and export them separately as well as a minor breaking change capitalization typo fix.

### Breaking Change:
There's one minor breaking change that seemed worth doing since this is a large enough change anyhow. Can revert if not worth it.
`MongoLogManager.cleanupOldLogfiles` => `MongoLogManager.cleanupOldLogFiles`

Related to: 
[MONGOSH-1992:](https://jira.mongodb.org/browse/MONGOSH-1992) Split MongoLogWriter and MongoLogManager
[MONGOSH-1991:](https://jira.mongodb.org/browse/MONGOSH-1991) Move logging to devtools-shared


packages/mongodb-log-writer/README.md
```diff
@@ -0,0 +1,3 @@
+# mongodb-constants
+
+Shared MongoDB constants for queries, aggregations, bson types, etc
```
addaleax, line 3: ^

packages/mongodb-log-writer/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@mongodb-js/mongodb-log-writer",
+  "description": "A library for writing MongoDB logv2 messages",
+  "version": "2.0.0",
```
addaleax, line 4: Heads up, especially since this is a major release, I'd include this in mongosh once this is done, then do a package release over there and update browser-repl & friends once that's done, so that all our products are on the 2.x version of this

packages/mongodb-log-writer/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "logv2"
+  ],
+  "main": "dist/index.js",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js",
+      "import": "./.esm-wrapper.mjs"
+    },
+    "./mongo-log-writer": "./dist/mongo-log-writer.js",
+    "./mongo-log-manager": "./dist/mongo-log-manager.js"
```
addaleax, line 33: I think you might also need `types` entries for the individual exports (as well as a top-level `types` property in this package.json), like e.g. https://github.com/mongodb/js-bson/blob/eca63c9f69225270f852f8097d73bb5b7540986f/package.json#L81

packages/mongodb-log-writer/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "lint": "eslint src/**/*.ts",
+    "test-only": "nyc mocha --colors -r ts-node/register src/**.spec.ts",
+    "test": "npm run lint && npm run build && npm run test-only",
+    "build": "npm run compile-ts && gen-esm-wrapper . ./.esm-wrapper.mjs",
+    "prepack": "npm run build",
+    "compile-ts": "tsc -p tsconfig.json"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "bson": "^4.5.1 || ^5.0.0 || ^6.0.0",
```
addaleax, line 52: If we're doing a major release anyway, maybe we can just stick to 6.x?

gagik: `depalign` ended up setting this to `^6.8.0` (highest devDependency version of bson in other repos), could that be too high for compatibility or is that fine?

addaleax: I don't really see how it would hurt :) 

Although it is honestly a bit annoying that depalign wants to manage this version for us. Maybe we can make this a `peerDependency`? Because the idea isn't to force bson updates on the users of this package with every new release

gagik: It technically doesn't manage it by default (I used `--autofix` and then rolled back most of the changes as it was doing too much, this one I just left). I could set it as `peerDepedency` with `6.x` then?

addaleax: Yeah, I'd do that 👍

packages/mongodb-log-writer/package.json
```diff
@@ -0,0 +1,90 @@
+{
+  "name": "@mongodb-js/mongodb-log-writer",
```
alenakhineika, line 2: Are we moving it under the `@mongodb-js/` prefix? originally the package had the `mongodb-log-writer` namespace. I think we have both namespaces in `devtools-shared`. Looks like the packages that was moved to the mono repo didn't add the `@mongodb-js/` prefix to their names. cc @addaleax

gagik: probably worth preserving the old name just for the NPM history etc.

PR #500, opened 2025-01-22
WIP feat!(mongodb-log-writer): add ability to disable logging MONGOSH-1988

gagik:
WIP - do not review

Adds a new `isDisabled` field to disable the writing of logs. This can be modified using the new `enable()` and `disable()` methods.

See: https://github.com/mongodb-js/mongosh/pull/2325

### DIFF
--- a/packages/mongodb-log-writer/src/mongo-log-writer.ts
+++ b/packages/mongodb-log-writer/src/mongo-log-writer.ts
@@ -63,6 +63,10 @@ function validateLogEntry(info: MongoLogEntry): Error | null {
   return null;
 }
 
+export type MongoLogWriterOptions = {
+  isDisabled?: boolean;
+};
+
 /**
  * A helper class for writing formatted log information to an output stream.
  * This class itself is an object-mode Writable stream to which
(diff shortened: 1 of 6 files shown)


packages/mongodb-log-writer/src/mongo-log-writer.ts
```diff
@@ -186,6 +212,13 @@ export class MongoLogWriter extends Writable {
     await new Promise((resolve) => this._target.write('', resolve));
   }
 
+  writeIfEnabled(chunk: Parameters<typeof this.write>[0]): void {
```
gagik, line 215: I could also use our current override of `this._write` and add the conditional there, it just feels less in our control so this feels more predictable to me.

addaleax:
The point of using something like `dummySink` from the tech design was to avoid maintaining this extra state here, did something change?

gagik:
@addaleax This is about to get revamped as I didn't originally follow the tech design, thus I put it into draft again

gagik:
Closing as there are no changes needed in the log writer package itself for the purposes of https://github.com/mongodb-js/mongosh/pull/2325

PR #504, opened 2025-02-06
feat(mongodb-log-writer): add `retentionGB` configuration MONGOSH-1985

gagik:
Adds `retentionGB` which enforces maximal GB of log files which are kept.

### DIFF
--- a/packages/mongodb-log-writer/src/mongo-log-manager.spec.ts
+++ b/packages/mongodb-log-writer/src/mongo-log-manager.spec.ts
@@ -1,6 +1,7 @@
 import { MongoLogManager, mongoLogId } from '.';
 import { ObjectId } from 'bson';
 import { once } from 'events';
+import type { Stats } from 'fs';
 import { promises as fs } from 'fs';
 import path from 'path';
 import os from 'os';
@@ -27,6 +28,7 @@ describe('MongoLogManager', function () {
   });
   afterEach(async function () {
     await fs.rmdir(directory, { recursive: true });
+    sinon.restore();
   });
 
   it('allows creating and writing to log files', async function () {
@@ -122,6 +124,58 @@ describe('MongoLogManager', function () {
     expect(await getFilesState(paths)).to.equal('1111111111');
     await manager.cleanupOldLogFiles();
     expect(await getFilesState(paths)).to.equal('0000011111');
+  });
+
+  it('if fs.stat fails, it errors and is not considered towards the logs limit', async function () {
+    const manager = new MongoLogManager({
+      directory,
+      retentionDays,
+      retentionGB: 3,
+      onwarn,
+      onerror,
+    });
+
+    const offset = Math.floor(Date.now() / 1000);
+
+    const faultyFile = path.join(
+      directory,
+      ObjectId.createFromTime(offset - 10).toHexString() + '_log'
+    );
+    await fs.writeFile(faultyFile, '');
+
+    const faultyFileError = new Error('test error');
+
+    const validFiles: string[] = [];
+    // Create 5 valid files.
+    for (let i = 5; i >= 0; i--) {
+      const filename = path.join(
+        directory,
+        ObjectId.createFromTime(offset - i).toHexString() + '_log'
+      );
+      await fs.writeFile(filename, '');
+      validFiles.push(filename);
+    }
+
+    expect(onerror).not.called;
+
+    const fsStatStub = sinon.stub(fs, 'stat');
+
+    fsStatStub.resolves({
+      size: 1024 * 1024 * 1024,
+    } as Stats);
+    fsStatStub.withArgs(faultyFile).rejects(faultyFileError);
+
+    await manager.cleanupOldLogFiles();
+
+    expect(onerror).calledOnceWithExactly(faultyFileError, faultyFile);
+
+    // fs.stat is stubbed so getFilesState will not be accurate.
+    const leftoverFiles = (await fs.readdir(directory))
+      .sort()
+      .map((file) => path.join(directory, file));
+
+    expect(leftoverFiles).to.have.lengthOf(4);
+    expect(leftoverFiles).deep.equals([faultyFile, ...validFiles.slice(3)]);
   });
 
   it('cleans up least recent log files when requested with a storage limit', async function () {
--- a/packages/mongodb-log-writer/src/mongo-log-manager.spec.ts
+++ b/packages/mongodb-log-writer/src/mongo-log-manager.spec.ts
@@ -86,6 +86,19 @@ describe('MongoLogManager', function () {
     }
   });
 
+  const getFilesState = async (paths: string[]) => {
+    return (
+      await Promise.all(
+        paths.map((path) =>
+          fs.stat(path).then(
+            () => 1,
+            () => 0
+          )
+        )
+      )
+    ).join('');
+  };
+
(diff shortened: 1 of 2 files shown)


packages/mongodb-log-writer/src/mongo-log-manager.spec.ts
```diff
@@ -205,7 +235,7 @@ describe('MongoLogManager', function () {
       .toString()
       .split('\n')
       .filter(Boolean)
-      .map((line: string) => JSON.parse(line));
+      retentionGBring) => JSON.parse(line));
```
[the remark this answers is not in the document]
gagik: really confused by how this happened 😮

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -69,23 +77,50 @@ export class MongoLogManager {
...
-        toDelete = fullPath;
-      } else if (this._options.maxLogFileCount) {
-        leastRecentFileHeap.push({ fullPath, fileTimestamp });
-        if (leastRecentFileHeap.size() > this._options.maxLogFileCount) {
-          toDelete = leastRecentFileHeap.pop()?.fullPath;
+        toDelete = {
+          fullPath,
+        };
+      } else if (this._options.retentionGB || this._options.maxLogFileCount) {
+        const fileSize = (await fs.stat(fullPath)).size;
```
addaleax, line 97: We should be prepared to handle an error from `fs.stat()` here

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -94,9 +94,17 @@ export class MongoLogManager {
...
+
         if (this._options.retentionGB) {
-          usedStorageSize += fileSize;
+          try {
+            fileSize = (await fs.stat(fullPath)).size;
+            if (this._options.retentionGB) {
+              usedStorageSize += fileSize;
+            }
+          } catch (err) {
+            this._options.onerror(err as Error, fullPath);
```
gagik, line 106: is this reasonable behavior or should we cancel the cleanup altogether in this case?

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -69,23 +77,58 @@ export class MongoLogManager {
...
+          fileSize,
+        });
+
+        const reachedMaxStorageSize = usedStorageSize > storageSizeLimit;
+        const reachedMaxFileCount =
+          this._options.maxLogFileCount &&
+          leastRecentFileHeap.size() > this._options.maxLogFileCount;
+
+        if (reachedMaxStorageSize || reachedMaxFileCount) {
+          toDelete = leastRecentFileHeap.pop();
```
addaleax, line 122: So ... took some time to think this through, but the problem with this approach is that we'll generally want to delete the _oldest_ files that are necessary to ensure that max file count + max file size limitations

But with this approach, that's no longer the case – you could end up with e.g. the most recent file being deleted, if it happens to be the first file we read from the directory and it's already over the size limit – right?

(I'm not sure if we need to let go of the heap as the underlying data structure, certainly wouldn't be the end of the world, but I'm pretty sure that the "read one file, check conditions, maybe delete one file" approach won't work anymore)

gagik: hmm yeah, we could sort by creation date (which would be same as the log's name?) from the getgo and adjust the logic accordingly

packages/mongodb-log-writer/src/mongo-log-manager.ts
addaleax: It seems very reasonable to me 👍

packages/mongodb-log-writer/src/mongo-log-manager.ts
addaleax: > which would be same as the log's name?

Yes, that's an assumption I'd still keep around 👍

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -38,9 +39,37 @@ export class MongoLogManager {
...
+      fullPath: string;
+      id: string;
+      size?: number;
+    }[] = [];
+    let usedStorageSize = this._options.retentionGB ? 0 : -Infinity;
+
     try {
-      dirHandle = await fs.opendir(dir);
+      const files = await fs.readdir(dir, { withFileTypes: true });
+      for (const file of files) {
```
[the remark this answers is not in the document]
gagik: oh yeah, I originally had a `reduce`-based logic (which is also how the sort got lost...) which is why I switched to read, will switch back

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -49,48 +78,56 @@ export class MongoLogManager {
...
+          fullPath,
+        };
+      } else if (this._options.retentionGB || this._options.maxLogFileCount) {
+        const reachedMaxStorageSize = usedStorageSize > storageSizeLimit;
+        const reachedMaxFileCount =
+          this._options.maxLogFileCount &&
+          sortedLogFiles.length > this._options.maxLogFileCount;
+
+        if (reachedMaxStorageSize || reachedMaxFileCount) {
+          toDelete = sortedLogFiles.shift();
```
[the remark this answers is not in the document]
gagik: I'm iterating on a shallow copy of it. We can either pop items from it or keep an index of deletion, I thought an index wouldn't be worth the readability loss but I guess array copying etc. isn't that much better.

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -38,9 +39,37 @@ export class MongoLogManager {
...
+
+        if (!file.isFile() || !id) {
+          continue;
+        }
+
+        const fullPath = path.join(dir, file.name);
+        let size: number | undefined;
+        if (this._options.retentionGB) {
+          try {
+            size = (await fs.stat(fullPath)).size;
```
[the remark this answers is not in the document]
gagik: not sure I understand. do you mean if a `stat` call fails for like 1 file or all files? if I follow right, I'd expect that specific file to be ignored and the rest of the logic would apply to the rest of the files

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -67,32 +84,57 @@ export class MongoLogManager {
...
+        leastRecentFileHeap.size() > this._options.maxLogFileCount
+      ) {
+        const toDelete = leastRecentFileHeap.pop();
+        if (!toDelete) continue;
+        await this.deleteFile(toDelete.fullPath);
+        usedStorageSize -= toDelete.fileSize ?? 0;
+      }
+    }
+
+    if (this._options.retentionGB) {
```
gagik, line 125: I think this should address most of the concerns.
Not sure what other issues might be with `fs.stat` worth handling

packages/mongodb-log-writer/src/mongo-log-manager.ts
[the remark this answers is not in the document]
addaleax: Yeah, so, the previous implementation would have a `stat()` call that would interrupt the entire method, that's gone now so all good 👍

addaleax (approved):

PR #505, opened 2025-02-06
feat(mongodb-log-writer): add ability to specify a prefix for log files

gagik:
This is needed for mongosh to use a `mongosh_` prefix for custom log directories.

Built on top of https://github.com/mongodb-js/devtools-shared/pull/504

### DIFF
--- a/packages/mongodb-log-writer/src/mongo-log-manager.ts
+++ b/packages/mongodb-log-writer/src/mongo-log-manager.ts
@@ -21,10 +21,10 @@ interface MongoLogOptions {
   retentionGB?: number;
   /** Prefix to use for the log files */
   prefix?: string;
-  /** A handler for warnings related to a specific filesystem path. */
-  onerror: (err: Error, path: string) => unknown | Promise<void>;
   /** A handler for errors related to a specific filesystem path. */
   onerror: (err: Error, path: string) => unknown | Promise<void>;
+  /** A handler for warnings related to a specific filesystem path. */
+  onwarn: (err: Error, path: string) => unknown | Promise<void>;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -73,8 +79,11 @@ export class MongoLogManager {
       if (Date.now() - deletionStartTimestamp > maxDurationMs) break;
 
       if (!dirent.isFile()) continue;
-      const { id } =
-        /^(?<id>[a-f0-9]{24})_log(\.gz)?$/i.exec(dirent.name)?.groups ?? {};
+      const logRegExp = new RegExp(
+        `^${this.prefix}(?<id>[a-f0-9]{24})_log(\\.gz)?$`,
```
addaleax, line 83: If we do insert the prefix into the regexp, we should either escape it or validate that it does not contain any characters that would have special meaning when interpreted as a regexp in the `MongoLogManager` constructor

gagik: added regex validation at point of setting the prefix that just limits it to alphanumeric and _

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -34,9 +36,20 @@ export class MongoLogManager {
   _options: MongoLogOptions;
 
   constructor(options: MongoLogOptions) {
+    if (options.prefix) {
+      if (!/^[a-z0-9_]+$/i.test(options.prefix)) {
```
addaleax, line 40: nit: Merging these two `if`s would make this a bit easier to read imo 🙂 

```suggestion
    if (options.prefix && !/^[a-z0-9_]+$/i.test(options.prefix)) {
```

addaleax (approved):

alenakhineika (approved):

PR #508, opened 2025-02-11
fix(ci): use npm@9.x to avoid failing silently

gagik:
Currently even if a lerna task fails, lerna still exits with success and so does the CI task.

The cause is the same as https://github.com/mongodb-js/mongosh/pull/1592 with NPM not sending the correct exit code. NPM 9 fixed this so this updates to that.

Example of a task which will actually fail now: https://github.com/mongodb-js/devtools-shared/actions/runs/13263792191/job/37026159185?pr=508

### DIFF
--- a/.github/workflows/check-test.yaml
+++ b/.github/workflows/check-test.yaml
@@ -14,6 +14,10 @@ on:
 permissions:
   contents: read # we just need to checkout the repo
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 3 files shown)


.github/workflows/check-test.yaml
```diff
@@ -14,6 +14,10 @@ on:
 permissions:
   contents: read # we just need to checkout the repo
 
+concurrency:
```
gagik, line 17: Drive-by improvement to cancel running tasks if a new commit is pushed

lerouxb (approved):

PR #515, opened 2025-03-05
feat(mongodb-log-writer): skip checks for `retentionGB` and `maxLogFileCount` when set to Infinity MONGOSH-2047

gagik:
Because in mongosh (and generally), some of our settings like `logRetentionGB` can be undefined while others like `maxLogFileCount` cannot, it's likely users will end up setting multiple of them as `Infinity` if they're looking to unset different things.

Our code skips some checks when some steps when those limits are `undefined` but it won't skip if they're infinite so it can be a performance improvement if we skip those steps if their values are infinite as well.

### DIFF
--- a/packages/mongodb-log-writer/src/mongo-log-manager.ts
+++ b/packages/mongodb-log-writer/src/mongo-log-manager.ts
@@ -62,7 +62,10 @@ export class MongoLogManager {
   }
 
   /** Clean up log files older than `retentionDays`. */
-  async cleanupOldLogFiles(maxDurationMs = 5_000, remainingRetries = 1): Promise<void> {
+  async cleanupOldLogFiles(
+    maxDurationMs = 5_000,
+    remainingRetries = 1
+  ): Promise<void> {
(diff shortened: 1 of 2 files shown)


packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -145,11 +155,14 @@ export class MongoLogManager {
...
-        await this.cleanupOldLogFiles(maxDurationMs - (Date.now() - deletionStartTimestamp), remainingRetries - 1);
+        await this.cleanupOldLogFiles(
+          maxDurationMs - (Date.now() - deletionStartTimestamp),
+          remainingRetries - 1
+        );
       }
     }
 
-    if (this._options.retentionGB) {
+    if (hasRetentionGB && this._options.retentionGB) {
```
gagik, line 165: I realize this is redundant but TypeScript isn't smart enough to do type promotion otherwise

addaleax (approved):

PR #517, opened 2025-03-06
feat: add ts-autocomplete package MONGOSH-2033

lerouxb:
This is extracted from the [PoC](https://github.com/mongodb-js/poc-mongodb-autocomplete). It doesn't know anything about what's in the TypeScript files, so should be generically useful.

In the end I added back support for filtering the completions, because while adding more tests I noticed there are just tons of cases where it returns the entire global scope or at least adds the entire global scope to the results. So I'm finding it hard to imagine us ever using this without at least some filtering of some kind.


packages/ts-autocomplete/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+To apply the Apache License to your work, attach the following
+boilerplate notice, with the fields enclosed by brackets "{}"
+replaced with your own identifying information. (Don't include
+the brackets!)  The text should be enclosed in the appropriate
+comment syntax for the file format. We also recommend that a
+file or class name and description of purpose be included on the
+same "printed page" as the copyright notice for easier
+identification within third-party archives.
+
+Copyright {yyyy} {name of copyright owner}
```
kmruiz, line 189: Totally-a-Lawyer

packages/ts-autocomplete/src/index.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    expect(autoCompleter.autocomplete('foo.', 3)).to.deep.equal([]);
+
+    // TODO: is this expected?
+    expect(autoCompleter.autocomplete('fo')).to.deep.equal([]);
+
+    // TODO: is this expected?
+    expect(autoCompleter.autocomplete('foo.b')).to.deep.equal([
+      { kind: 'property', name: 'a', type: 'string' },
+      { kind: 'property', name: 'bb', type: 'function (p1: number)' },
+    ]);
```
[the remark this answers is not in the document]
lerouxb: Where should we be calculating that priority?

packages/ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    const completions = this.languageService.getCompletionsAtPosition(
+      'main.ts',
+      position,
+      {
+        allowIncompleteCompletions: true,
+        includeSymbol: true,
+      }
+    );
+
+    if (completions?.isMemberCompletion) {
```
kmruiz, line 100: IIRC, this basically filters out any AC that is not a "member" of another. Basically, it only autocompletes things that are:

`a.|`

where | is the caret position.

packages/ts-autocomplete/src/index.spec.ts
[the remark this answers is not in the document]
addaleax: I'd have expected the consumer to do this completion, e.g. in mongosh we'd only complete proper prefixes whereas the VSCode GUI shows a list of fuzzy-matched options.

@kmruiz Is that something the products do that talk to the language server there, or is it something the language server is expected to do? I think that would influence our decision here quite a bit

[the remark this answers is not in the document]
kmruiz: It's the product who actually does the prioritisation. It's done this way because there are different ways to sort and the product is free to implement their own.

I guess for mongosh, the approach you mention @addaleax is the best one. For Compass I guess it would depend on how CodeMirror works, and for vscode we don't need to filter nor sort at all.

packages/ts-autocomplete/src/index.ts
lerouxb: yeah. Didn't notice at the time and now it is obvious in hindsight. The trouble with copy/pasting someone else's code, I guess 😆

packages/ts-autocomplete/LICENSE
lerouxb: Screenshot 2025-03-06 at 15 08 54

🤷

packages/ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+function mapCompletions(completions: ts.CompletionInfo) {
+  return completions.entries.map((entry) => {
+    const declarations = entry.symbol?.getDeclarations();
+    let type = 'any';
+    if (declarations && declarations[0]) {
+      const decl = declarations[0];
+      type = decl.getChildAt(2)?.getFullText()?.trim() ?? 'any';
+    }
```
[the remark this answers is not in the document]
lerouxb: Looking into it. Also: why would we use only the first declaration?

packages/ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,188 @@
...
+    .map((entry) => {
+      // entry.symbol is included because we specify includeSymbol when calling
+      // getCompletionsAtPosition
+      const declarations = entry.symbol?.getDeclarations();
+      let type = 'any';
+      const decl = declarations?.[0];
+      if (decl) {
+        // decl's children are things like ['a', ':', 'string'] or ['bb', ':',
+        // '(p1: number) => void']. So the one at position 2 (zero indexed) is the
+        // type.
```
addaleax, line 127: 💙

addaleax (approved):

PR #520, opened 2025-03-19
feat: add mongodb-ts-autocomplete package MONGOSH-2034

lerouxb:
Currently still works with fake mql  types, but the shell-api ones are at least the real ones.

**TODO: separate tickets**
* use real mql.ts [MONGOSH-2032](https://jira.mongodb.org/browse/MONGOSH-2032)


packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,279 @@
...
+    // in combination with either the shell-api or something like a document or
+    // aggregation editor
+
+    this.autocompleter.updateCode({
+      'bson-expressions.d.ts': loadBSONExpressions(),
+    });
+    this.autocompleter.updateCode({ 'shell-api.d.ts': loadShellAPI() });
+  }
+
+  setConnectionKey(connectionKey: string) {
```
lerouxb, line 144: I have this idea of setting a connectionKey (which should probably be renamed to connectionId) and databaseName separate from the autocomplete() function, but I think a better API would be to have the autocomplete method just take the connectionId and databaseName as parameters. Then we don't need the concept of which connection and/or database is the current one.

Every time we autocomplete we already re-build current-globals.d.ts anyway because the current collection schema could have changed, so it might as well take that into account right there. We already have a transparent cache for the schemas regardless and the server schema is calculated and updated as one chunk.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,279 @@
...
+    const databaseName = this.getActiveConnection().getCurrentDatabaseName();
+    activeConnection.addCollectionSchema(databaseName, collectionName, schema);
+
+    const collectionNames = await this.context.collectionsForDatabase(
+      this.currentConnectionKey,
+      databaseName
+    );
+    activeConnection.setDatabaseCollectionNames(databaseName, collectionNames);
+
+    // TODO: the problem with doing it this way is that, while the collection
```
lerouxb, line 263: I think this is probably fine. Getting the sample docs and calculating a schema is what's relatively expensive, probably not walking it and writing a few lines of typescript. Only way this becomes an issue is if you have a really large collection schema or very many databases/collections in the same server schema. Each individual database's list of collections is lazy loaded, though and also each collection's schema. So you'd likely have to have quite the long lived and meandering session for that to build up.

We would likely load the list of databases or collections for at least one database early on, though.

packages/mongodb-ts-autocomplete/src/index.ts
lerouxb: Ok this was dumb - cleaned up now.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,256 @@
...
+  private collectionNames: Set<string>;
+  private collectionSchemas: Record<string, JSONSchema>;
+
+  constructor() {
+    // TODO: this is kinda the only real reason for this class: So we can keep
+    // track of the known collectionNames for a database. It could be all the
+    // names from a listCollections() or it could just be all the ones we've
+    // auto-completed for. The schemas can come from the autocompletion context.
+    // This can probably be added to the autocompletion context.
+    this.collectionNames = new Set();
```
lerouxb, line 38: Thinking about whether DatabaseSchema and ConnectionSchema can just be merged into the autocompletion context and then it keeps track of the known database names and for each database the known collection names.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,279 @@
...
+
+    this.connectionSchemas = Object.create(null);
+    this.connectionSchemas;
+
+    // TODO: set the @mongodb-js/mql-typescript definitions (will this be separate or part of shell-api?)
+    // these aren't globals because they are types, so probably only make sense
+    // in combination with either the shell-api or something like a document or
+    // aggregation editor
+
+    this.autocompleter.updateCode({
```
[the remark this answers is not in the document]
lerouxb: OK the plan is basically to build modules we can import that just export one large string that's the typescript definitions we have to use in the autocomplete language service.

So basically:
mql.d.ts, bson.d.ts, bson-expressions.d.ts, shell-api.d.ts all get into the language service that way.

Then we reference them like in the files:
```
/// <reference path="bson" />
/// <reference path="mql" />
/// <reference path="bson-expressions" />
/// <reference path="shell-api" />
```

mql, bson and shell-api would all be namespaced. Not sure about bson-expressions: I think those are just global.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,279 @@
...
+  ): Promise<AutoCompletion[]> {
+    if (!this.currentConnectionKey) {
+      throw new Error('No active connection');
+    }
+
+    if (typeof position === 'undefined') {
+      position = code.length;
+    }
+
+    // TODO: we're now compiling the source twice: autocomplete will also
```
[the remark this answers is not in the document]
lerouxb: (but this would also totally be a micro-optimisation at this stage)

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  getCurrentGlobalsCode(connectionId: string, databaseName: string) {
+    return `
+import { ServerSchema } from '/${connectionId}.ts';
+
+type CurrentDatabaseSchema = ServerSchema['${databaseName}'];
+
+declare global {
+  var db: CurrentDatabaseSchema;
+  var use: (collection: string) => CurrentDatabaseSchema;
```
nirinchev, line 163: Hm... is this the `use db` function or something else? Why does it take a collection and return schema?

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+    // other collection in the db and in fact every db in the server) every
+    // time.
+    this.autocompleter.updateCode({
+      [`/${connectionId}.ts`]: this.getConnectionCode(connectionId),
+    });
+    this.autocompleter.updateCode({
+      '/current-globals.ts': this.getCurrentGlobalsCode(
+        connectionId,
+        databaseName
+      ),
```
nirinchev, line 210: We can probably cache those quite easily? I agree that it might not be a problem, but it should be fairly low effort since we already have the caches.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+    }
+    return this.connectionSchemas[connectionId];
+  }
+
+  getConnectionCode(connectionKey: string): string {
+    return `
+import * as ShellAPI from '/shell-api.ts';
+import * as bson from '/bson.ts';
+
+export type ServerSchema = ${this.connectionSchemas[
```
nirinchev, line 149: Do we know if the compiler behaves differently depending on whether we dump all of that in a huge file vs generate multiple small files instead? We could experiment with generating a file per collection and then importing all of those in the connection file. That way, if we need to add a new collection, we'd only need to regenerate a subset of the entire code.

nirinchev (commented):
Haven't gone through all of it yet, but leaving some early thoughts - most are nits, but there are a few actual concerns. Will try to get through the rest later today or on Monday.

packages/mongodb-ts-autocomplete/src/index.ts
lerouxb: The problem then is that we have to figure out how to convert from valid mongodb collection names to valid typescript type names. The benefit of having a giant per-server schema is that we can address it with `['database or collection name']`.

I started having one per collection and per database and per connection, then simplified it like this when the idea came from a thread in slack: https://mongodb.slack.com/archives/C08EXTB42LB/p1742298506300629

But I'm not convinced about this to the point that I can't be persuaded either way :)

packages/ts-autocomplete/src/index.ts
```diff
@@ -102,12 +113,27 @@ function getSymbolAtPosition(
...
+export type AutoCompletion = {
   name: string;
   kind: ts.ScriptElementKind;
   type: string;
 };
 
+function getTypeFromDeclaration(decl: ts.Declaration): string {
+  const index = decl
+    .getChildren()
+    .findIndex((child) => child.getFullText() === ':');
```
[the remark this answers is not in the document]
lerouxb: I'm just removing this for now as type won't get used and it doesn't work well. Won't work well without quite a bit of work.

packages/ts-autocomplete/src/index.ts
```diff
@@ -145,18 +174,30 @@ type AutocompleteFilterFunction = (
   filterOptions: AutocompleteFilterOptions
 ) => boolean;
 
-type AutocompleterOptions = {
+export type AutocompleterOptions = {
   filter?: AutocompleteFilterFunction;
 };
 
+function filterDiagnostics(diagnostics: any[]) {
```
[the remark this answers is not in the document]
lerouxb: I agree with that rule, I just didn't bother because it was debug info. Sorry about that - fixed now.

packages/ts-autocomplete/src/index.ts
```diff
@@ -145,18 +174,30 @@ type AutocompleteFilterFunction = (
...
   private readonly filter: AutocompleteFilterFunction;
   private readonly languageService: ts.LanguageService;
   readonly updateCode: UpdateDefinitionFunction;
+  readonly listfiles: () => string[];
 
   constructor({ filter }: AutocompleterOptions = {}) {
     this.filter = filter ?? (() => true);
-    [this.languageService, this.updateCode] = getVirtualLanguageService();
+    [this.languageService, this.updateCode, this.listfiles] =
+      getVirtualLanguageService();
```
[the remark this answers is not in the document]
lerouxb: I agree. I'll change it. I just copy/pasted it like that from the PoC and never updated it.

packages/mongodb-ts-autocomplete/src/type-export.ts
```diff
@@ -0,0 +1,162 @@
...
+
+function uniqueTypes(property: JSONSchema): Set<string> {
+  const type = getBSONType(property);
+  const types = Array.isArray(type)
+    ? type.map((t) => toTypeName(t))
+    : [toTypeName(type ?? 'any')];
+  return new Set(types.flat());
+}
+
+function indentSpaces(indent: number) {
```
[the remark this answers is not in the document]
lerouxb: No, but it becomes a giant block that I want to be able to read in test output and fixtures.

packages/mongodb-ts-autocomplete/src/type-export.ts
```diff
@@ -0,0 +1,162 @@
...
+    spaces.push('  ');
+  }
+  return spaces.join('');
+}
+
+function arrayType(types: string[]) {
+  if (types.length === 1) {
+    return `${types[0]}[]`;
+  }
+  return `${types.join(' | ')})[]`;
```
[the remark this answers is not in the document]
lerouxb: See this is why its handy to have the output formatted and then look at it as one block in a test fixture 😄

packages/mongodb-ts-autocomplete/src/type-export.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  return 'any';
+}
+
+function uniqueTypes(property: JSONSchema): Set<string> {
+  const type = getBSONType(property);
+  const types = Array.isArray(type)
+    ? type.map((t) => toTypeName(t))
+    : [toTypeName(type ?? 'any')];
+  return new Set(types.flat());
```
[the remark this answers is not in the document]
lerouxb: nice.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+import type { AutocompletionContext } from './autocompletion-context';
+
+type MongoDBAutocompleterOptions = {
+  context: AutocompletionContext;
+  autocompleterOptions?: AutocompleterOptions;
+};
+
+class DatabaseSchema {
+  private collectionNames: Set<string>;
+  private collectionSchemas: Record<string, JSONSchema>;
```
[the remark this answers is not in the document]
lerouxb: Yeah I'll do that, but my concern around ConnectionSchema and DatabaseSchema relates more to whether I should have those objects at all given that we already have a cache for anything we could get via the autocompletion context.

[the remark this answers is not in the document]
lerouxb: I made the change.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    return `{
+  ${collectionProperties.join('\n')}
+}`;
+  }
+}
+
+class ConnectionSchema {
+  private readonly databaseNames: Set<string>;
+  private readonly databaseSchemas: Record<string, DatabaseSchema>;
```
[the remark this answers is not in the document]
lerouxb: Made the change, but same comment as above.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  constructor() {
+    // TODO: this is kinda the only real reason for this class: So we can keep
```
lerouxb, line 78: I'll remove the TODO before merging, just keen on hearing people's thoughts first.

packages/mongodb-ts-autocomplete/src/index.ts
```diff
@@ -0,0 +1,209 @@
...
+type MongoDBAutocompleterOptions = {
+  context: AutocompletionContext;
+  autocompleterOptions?: AutocompleterOptions;
+};
+
+class DatabaseSchema {
+  private collectionSchemas: Record<string, JSONSchema | undefined>;
+
+  constructor() {
+    // TODO: this is kinda the only real reason for this class: So we can keep
```
lerouxb, line 31: I'll remove the TODO before merging, just keen on hearing people's thoughts first.

packages/mongodb-ts-autocomplete/src/index.ts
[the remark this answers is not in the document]
lerouxb: Again - I'll just remove the TODO before merging. Just thought I'd point this out to reviewers. I think this is OK since it is just a line or small multi-line snippet of code and triggering autocomplete is not exactly something that happens in a tight loop. We can optimise it later if we ever have to.

Just wanted to point out that I did think about it  😆

packages/mongodb-ts-autocomplete/src/index.ts
lerouxb: Similar to comments above, I think I'll remove this comment before merging. Just wanted to point out that I'm aware.

packages/mongodb-ts-autocomplete/scripts/extract-types.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable no-console */
+import { promises as fs } from 'fs';
+import path from 'path';
+
+function replaceImports(code: string) {
+  // This just makes it possible to work on mql.ts because then the
```
lerouxb, line 6: Actually my stubbed mql.ts file doesn't even use the bson library at the moment, but this will become relevant again once we have the real mql.ts or otherwise use something that depends on loading in the bson library.

packages/mongodb-ts-autocomplete/src/autocompleter.ts
```diff
@@ -0,0 +1,230 @@
...
+  setCollectionSchema(collectionName: string, schema: JSONSchema): void {
+    this.collectionSchemas[collectionName] = schema;
+  }
+
+  toTypescriptTypeDefinition(): string {
+    const collectionProperties = Object.entries(this.collectionSchemas).map(
+      ([collectionName, schema]) => {
+        const def: string = schema ? toTypescriptTypeDefinition(schema) : `{}`;
+        const lines = def.split(/\n/g).map((line) => `      ${line}`);
+        return `  '${collectionName}': {
```
addaleax, line 53: Keep in mind that collection names can contain quotes, newlines, etc.

Should probably be
```suggestion
        return `  ${JSON.stringify(collectionName)}: {
```
?

packages/mongodb-ts-autocomplete/src/autocompleter.ts
```diff
@@ -0,0 +1,230 @@
...
+      collectionName,
+      collectionSchema,
+    );
+  }
+
+  toTypescriptTypeDefinition(): string {
+    const databaseProperties = Object.entries(this.databaseSchemas).map(
+      ([databaseName, schema]) => {
+        const def = schema.toTypescriptTypeDefinition();
+        return `'${databaseName}': ${def}`;
```
addaleax, line 99: ditto?

```suggestion
        return `${JSON.stringify(databaseName)}: ${def}`;
```

I mean, very generally you always want to `JSON.stringify()` when inserting JS values into JS source code, and then if you need to properly handle `__proto__` you might even want to do `JSON.parse(${JSON.stringify(JSON.stringify())})`

packages/mongodb-ts-autocomplete/src/index.ts
lerouxb: I filed a follow-up ticket for this https://jira.mongodb.org/browse/MONGOSH-2209

packages/mongodb-ts-autocomplete/src/autocompleter.ts
lerouxb: Makes sense. I'll change it throughout.

packages/mongodb-ts-autocomplete/src/autocompleter.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+        name: 'findOneAndUpdate',
+        result: 'db.foo.findOneAndUpdate',
+      },
+    ]);
+  });
+
+  // TODO: We need MONGOSH-2170 so that we can use the generated MQL types via
+  // the Shell API to autocomplete fields in
+  // ServerSchema[databaseName][collectionName].schema
+  it.skip('completes a collection field name in a query', async function () {
```
lerouxb, line 146: ☝️ Just FYI.

packages/mongodb-ts-autocomplete/src/fixtures/autocomplete-types.ts
addaleax, line 1: Should this file be checked into the repo?

addaleax (approved):

packages/mongodb-ts-autocomplete/scripts/extract-types.ts
```diff
@@ -0,0 +1,51 @@
...
+  `;
+
+  const filepath = path.join(
+    __dirname,
+    '..',
+    'src',
+    'fixtures',
+    'autocomplete-types.ts',
+  );
+  await fs.writeFile(filepath, code, 'utf-8');
```
lerouxb, line 41: Hmm. So the problem with doing it this way is that the shell api is now included in the bundle twice. Once in the api export which contains a giant `api` string and then again on this side where we generate a file to be included.

But I do kinda want to post-process the api here to change imports so unless we do the replaceImports() thing inside shell-api already there's probably no easy way around it.

Until we do the same thing for mql as we do for shell-api we kinda have to read the file as a string on this side and include the result in the bundle anyway so that we can pass that to the language service.

And for bson where we need the library as a string we have to read it at compile time and generate a string to be included in the bundle that can be passed to the language service regardless, because we're probably not going to be modifying the bson library to add that export.

packages/mongodb-ts-autocomplete/src/fixtures/autocomplete-types.ts
lerouxb: I removed it.

PR #521, opened 2025-03-20
feat(autocomplete): add type generator from the php yaml definitions

nirinchev:
## Description
Adds generators for the aggregation stages using the yaml definitions from https://github.com/mongodb-js/mongo-php-library/tree/v1.x/generator/config.

Three generators are added to process the yaml definitions:
1. `schema-generator.ts` generates a typescript schema from the yaml definitions which can be used to autocomplete aggregation stages. Notable limitations of the current schema include:
	- Unable to autocomplete nested fields e.g. `$unset: [ 'foo.bar' ]`
	- Stages following `$addFields`, `$project`, etc. will not reflect the new collection schema
	- Certain stages add variables which are not visible in the autocompletion (e.g. `$lookup.let`)
	- Some types don't have static schema defined (e.g. `SearchScore`, `GeoPoint`, etc.)
2. `driver-schema-generator.ts` generates the schema for the tests in the php yaml library. It does this by infering the schema from the documentation.
3. `test-generator.ts` generates tests from the yaml definitions. Those are not functional tests, but rather verify that the code compiles - that is, the schema definition does not generate false negatives. They don't attempt to verify that the autocompletion always suggests precise results though - i.e. a schema definition where everything is `any` would pass all the tests in this PR. Validating autocomplete results are helpful is a follow-up task that will be done when we integrate all the autocomplete tickets together.

## Open Questions
We probably want to get https://github.com/mongodb/mongo-php-library/pull/1659 eventually merged to ensure we don't diverge too far from upstream.


packages/mql-typescript/src/driverSchema/docsCrawler.ts
```diff
@@ -58,7 +58,7 @@ class IsoDateProcessor extends RegexCustomTypeProcessor {
         s || '00'
       }.${ms || '000'}${tz || 'Z'}`;
       const date = new Date(normalized);
-      // Make sur we're in the range 0000-01-01T00:00:00.000Z - 9999-12-31T23:59:59.999Z
+      // Make surd we're in the range 0000-01-01T00:00:00.000Z - 9999-12-31T23:59:59.999Z
```
lerouxb, line 61: Funny that you even tried fixing the typo 😛

packages/mql-typescript/src/driverSchema/docsCrawler.ts
```diff
@@ -0,0 +1,387 @@
...
+        date.getTime() >= -62167219200000 &&
+        date.getTime() <= 253402300799999
+      ) {
+        return date;
+      }
+    }
+
+    return new Date(value);
+  }
+}
```
addaleax, line 72: Yeah, this looks a lot like the type of logic we have in https://github.com/mongodb-js/devtools-shared/tree/main/packages/ejson-shell-parser 🙂

packages/mql-typescript/src/generator.ts
```diff
@@ -0,0 +1,292 @@
...
+              yield { yaml: parsed, path: filePath };
+            }
+          }
+        },
+      };
+    }
+  }
+
+  protected emitToFile(filePath: string): void {
+    this.outputStream = createWriteStream(filePath, { encoding: 'utf8' });
```
lerouxb, line 236: Must we use streams? I find it kinda redundant in most cases nowadays with async/await, async iterators, etc. It makes it tricky to try and do multiple things with the data at once. For example I want to also write the types we generate as a giant string elsewhere and streams force all the file writing into the whole structure of this class, making it tricky to just do two things or do one thing after another without restructuring it all.

addaleax (commented):
Amazing work @nirinchev! ✨

packages/mql-typescript/out/schema.ts
```diff
@@ -0,0 +1,6169 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable @typescript-eslint/no-namespace */
+/* eslint-disable @typescript-eslint/ban-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import type * as bson from 'bson';
+import type { FilterOperators } from 'mongodb';
```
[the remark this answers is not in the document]
lerouxb: Actually I just thought about this some more and I don't think we'll be including mql directly in mongodb-ts-autocomplete. We'll use it in the shell-api and probably just instruct api-extractor to include the relevant types because in the mongosh case that's the only way we'll be consuming it, same as for the bson expressions currently.

Later when we add support for other editors like a document editor or filter editor (in the compass querybar) or an aggregation editor we'll probably make something that bundles that up with whatever other shell syntax stuff we need (bson expressions?) and then re-export it from there.

So I think we don't even need "mql exported as a giant string" and there's no need to worry about importing bson and/or mongodb because that will just be handled downstream.

lerouxb (approved):
Great work! Like I explained I think we probably don't need https://github.com/mongodb-js/devtools-shared/pull/521/commits/7fc9890be30bf675bbb5c28e66493e4882ca43e3 anymore. Up to you if you want to remove it again.

packages/mql-typescript/package.json
```diff
@@ -0,0 +1,87 @@
...
+  "files": [
+    "dist",
+    "bin"
+  ],
+  "bin": {
+    "mql-typescript": "bin/runner.js"
+  },
+  "license": "Apache-2.0",
+  "main": "dist/index.js",
+  "exports": {
```
lerouxb, line 30: I think we need an export of schema.ts here somewhere so that we can import it from shell-api. Similar to what I started doing [here](https://github.com/mongodb-js/devtools-shared/commit/d7112f2e4894b34a82174461692fc0c167ede656) and what we do in shell api [here](https://github.com/mongodb-js/mongosh/blob/main/packages/shell-api/package.json#L12-L15). 

(Except the real thing so it can be imported, not the giant string.)

PR #523, opened 2025-03-28
chore(ci): run all checks always

nirinchev:
## Description
Changes the CI workflows to always run all tests, not just changed ones.

## Checklist

### DIFF
--- a/.github/workflows/check-test.yaml
+++ b/.github/workflows/check-test.yaml
@@ -99,9 +99,9 @@ jobs:
         # We exclude dependents when running eslint and depchecks because
         # linting and dependency checking is relevant only for packages where
         # the source code was changed
-        run: npm run check-ci -- --stream --since ${SINCE_REF} --exclude-dependents
+        run: npm run check-ci -- --stream
         shell: bash
 
       - name: Run Tests
-        run: npm run test-ci -- --stream --since ${SINCE_REF}
+        run: npm run test-ci -- --stream
         shell: bash


packages/node-webpack-startup-snapshot-checker/src/index.spec.ts
```diff
@@ -1,25 +1,52 @@
...
 import path from 'path';
 import { expect } from 'chai';
-const execFile = promisify(execFileCallback);
 
 describe('startup snapshot checker', function () {
+  const execFile = (
+    command: string,
+    args: ReadonlyArray<string>
+  ): Promise<void> => {
+    const proc = spawn(command, args);
```
nirinchev, line 10: Using `execFile` from `child_process` was causing `ENOENT` trying to run `npx` on windows.

packages/shell-bson-parser/src/index.spec.ts
```diff
@@ -448,51 +448,137 @@ describe('@mongodb-js/shell-bson-parser', function () {
           toISOString: (${newDate}).toISOString(),
           valueOf: (${newDate}.valueOf()),
        }`;
-              expect(parse(input, options)).to.deep.equal({
```
nirinchev, line 451: Had to rework this test due to race conditions where the test run could span multiple milliseconds, resulting in failures when comparing the fields that have millisecond precision.

addaleax (approved):

PR #524, opened 2025-03-31
chore(ci): enable coveralls MONGOSH-2125

nirinchev:
## Description
Enables coveralls reporting. Example run: https://coveralls.io/builds/72998592.

Waiting on IT to get admin access to the repo on Coveralls and fix the monorepo settings.

## Checklist


.github/workflows/check-test.yaml
```diff
@@ -95,12 +87,26 @@ jobs:
...
-        # linting and dependency checking is relevant only for packages where
-        # the source code was changed
         run: npm run check-ci -- --stream
         shell: bash
 
       - name: Run Tests
         run: npm run test-ci -- --stream
         shell: bash
+
+      - name: Report Coverage
```
gagik, line 97: fyi: coveralls does have an official github action that could be worth considering, didn't look too much into it though, might not fit our setup: https://github.com/marketplace/actions/coveralls-github-action

gagik (approved):

.github/workflows/check-test.yaml
nirinchev: Yeah, I looked at it, but the problem with it is that it's designed to report the results of a single package run, so I'd have to either add 20 steps to report coverage from each of the packages in the monorepo, or completely change how we're running CI to make it run one job per package in a matrix. The official action is using the CLI like we are and I did consult with its implementation when working on our reporting.

PR #525, opened 2025-04-01
feat(dl-center): add support for arbitrary acls COMPASS-9237

drichmdb:
This commit adjusts our DownloadCenter's UploadAssetOptions to support taking an optional 'acl' parameter. When we upload an asset to S3, this parameter is used to set the ACL on the put request. If no acl is set, we default to "public-read" to preserve backward compatibility.

### DIFF
--- a/packages/dl-center/src/download-center.ts
+++ b/packages/dl-center/src/download-center.ts
@@ -52,6 +52,7 @@ export type S3BucketConfig = {
 
 export type UploadAssetOptions = {
   contentType?: string;
+  acl?: string;
 };
 
 type S3UploadFunc = (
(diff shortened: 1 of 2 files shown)


packages/dl-center/src/download-center.ts
```diff
@@ -235,8 +236,10 @@ export class DownloadCenter {
       throw new Error('s3ObjectKey is required');
     }
 
+    const acl = options.acl === undefined ? ACL_PUBLIC_READ : options.acl;
```
lerouxb: `const acl = options.acl ?? ACL_PUBLIC_READ;` should work and be more modern/idiomatic.

lerouxb: nitpick, obviously. Feel free to ignore ;)

lerouxb (approved):

packages/dl-center/src/download-center.ts
drichmdb: Done!

PR #526, opened 2025-04-08
fix(mongodb-cloud-info): if the host's IP is not in a cloud's range, check if the CNAME resolves to AWS. COMPASS-8932

lerouxb:
This is to deal with the fact that Atlas has some of their own IP addresses for AWS hosts and that messes up telemetry reporting.

### DIFF
--- a/packages/mongodb-cloud-info/src/index.ts
+++ b/packages/mongodb-cloud-info/src/index.ts
@@ -22,6 +22,7 @@ export type RawCloudProviderCIDRs = {
 let unparsedCIDRsPromise: Promise<RawCloudProviderCIDRs> | undefined;
 
 const dnsLookup = util.promisify(dns.lookup.bind(dns));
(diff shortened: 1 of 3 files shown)


packages/mongodb-cloud-info/src/index.ts
```diff
@@ -41,6 +42,18 @@ function parseCIDRs(rawCidrs: RawCIDRs): ParsedCIDRs {
   };
 }
 
+async function hasAWSCname(host: string) {
+  try {
+    const addresses = await dnsResolveCname(host);
+    return addresses.some((address) => address.endsWith('.amazonaws.com'));
+  } catch (err: unknown) {
+    // This can be any of a long list of codes, but in all cases we're just
+    // going to assume that it is not one an AWS host.
```
addaleax, line 51:
```suggestion
    // going to assume that it is not on an AWS host.
```
?

addaleax (approved):

PR #527, opened 2025-04-14
feat: add a native machine ID library COMPASS-8443

gagik:
Adds a native device ID library for MacOS, Linux, and Windows. 100x+ faster implementation without depending on child process spawning.

Meant as a replacement for [node-machine-id](https://www.npmjs.com/package/node-machine-id) and since this is a popular library and usecase, I also put some effort to make this package more tidy and useful for the larger community.


packages/machine-id/scripts/benchmark.ts
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env node
+/* eslint-disable no-console */
```
gagik, line 2: This is more for fun, I was curious what the difference would be

packages/machine-id/binding.cc
```diff
@@ -0,0 +1,170 @@
...
+    {
+      return "";
+    }
+
+    CFStringRef uuidKey = CFSTR("IOPlatformUUID");
+    CFTypeRef uuidProperty = IORegistryEntryCreateCFProperty(ioRegistryRoot, uuidKey, kCFAllocatorDefault, 0);
+
+    if (uuidProperty)
+    {
+      char buffer[128];
```
gagik, line 36: maybe there's some buffer overflow possibility here? not sure

packages/machine-id/binding.cc
```diff
@@ -0,0 +1,170 @@
...
+  // Trim whitespace and newlines from a string
+  std::string trim(const std::string &str)
+  {
+    if (str.empty())
+    {
+      return str;
+    }
+
+    std::string result = str;
+    result.erase(result.find_last_not_of(" \n\r\t") + 1);
```
gagik, line 61: not sure if this is problematic / could end up in a weird state

packages/machine-id/src/index.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    it('returns a valid SHA256 hash format machine ID', function () {
+      // SHA256 hash format: 64 hex digits
+      const hashRegex = /^[0-9a-f]{64}$/i;
+
+      expect(hashRegex.test(id));
+
+      expect(id).equals(
+        createHash('sha256')
+          .update(getMachineId({ raw: true }) || '')
```
kraenhansen, line 64: Don't you want to assert this is never empty instead of assigning a default empty string as the value?

packages/machine-id/src/index.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      expect(hashedId).to.equal(nodeId);
+    });
+  });
+
+  describe('with hashing', function () {
+    let id: string;
+
+    beforeEach(function () {
+      id = getMachineId() || '';
```
kraenhansen, line 53: Don't you want to assert this is never empty instead of assigning a default empty string as the value?

packages/machine-id/src/index.spec.ts
kraenhansen, line 64: Same as above - why not assert a non-empty string?

packages/machine-id/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  | string
+  | undefined {
+  let machineId: string | undefined;
+
+  try {
+    machineId = binding.getMachineId();
+  } catch {
+    // If the binding fails, we can assume the machine ID is not available.
+    return undefined;
+  }
```
kraenhansen, line 25: I would likely put this into its own function to avoid the mutable `machineId`.

packages/machine-id/src/index.spec.ts
gagik: my general thinking was that it'd fail otherwise but fair point especially combined with not asserting `id` before.

packages/machine-id/binding.cc
gagik: good catch, I'll explicitly handle that case just to make the behavior more intentionally documented

kmruiz (changes requested):

kmruiz (approved):
Leaving it approved so you can merge when the comments have been resolved.

Great job!

packages/machine-id/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Get the machine ID for the current system
+ * @returns The machine ID (UUID) or undefined if not available
+ */
+export function getMachineId({ raw = false }: GetMachineIdOptions = {}):
+  | string
+  | undefined {
+  const machineId: string | undefined = getMachineIdFromBinding();
```
kraenhansen, line 27: This type seem completely inferable?

```suggestion
  const machineId = getMachineIdFromBinding();
```

kraenhansen (approved):

packages/machine-id/binding.cc
```diff
@@ -0,0 +1,173 @@
...
+  // Get Windows machine ID from registry
+  std::string getMachineId()
+  {
+    std::string machineGuid;
+    HKEY hKey;
+    LONG result = RegOpenKeyExA(
+        HKEY_LOCAL_MACHINE,
+        "SOFTWARE\\Microsoft\\Cryptography",
+        0,
+        KEY_QUERY_VALUE | KEY_WOW64_64KEY,
```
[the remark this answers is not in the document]
gagik: so one thing I forgot mention is that one goal of it this is to match Atlas CLI's (Go-based) device ID library so this copies the implementation from https://github.com/denisbrodbeck/machineid/blob/master/id_windows.go which does do this, unless `|` is different in C++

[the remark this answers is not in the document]
kmruiz: Yeah the operator does the same, the only thing is that this logic can break old 32 bit machines and don't do anything on 64 bit machines.

However if we want to do the same thing _just in case_ I'm fine keeping this logic as it is.

PR #531, opened 2025-05-02
[WIP DO NOT MERGE] Autocomplete2 demo

lerouxb:
This is these branches merged together to do an end to end in progress demo in Mongosh:

* https://github.com/mongodb-js/devtools-shared/pull/520
* https://github.com/mongodb-js/devtools-shared/pull/521

I have also added the `rs`, `sh` and `sp` globals which I'll have to backport.

**TODO:**

* [ ] Modify shell-api to use the mql types. Then hopefully we can start to autocomplete known collection field names.
* [ ] maybe the filtering/sorting of autocompletion results should happen here and not in mongosh. That way we can also make it smarter without breaking interface changes that cross repos.
* [ ] also try and add the method comments to the autocomplete result. We probably won't display it in mongosh yet, but we're more likely to in compass.


packages/mongodb-ts-autocomplete/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+To apply the Apache License to your work, attach the following
+boilerplate notice, with the fields enclosed by brackets "{}"
+replaced with your own identifying information. (Don't include
+the brackets!)  The text should be enclosed in the appropriate
+comment syntax for the file format. We also recommend that a
+file or class name and description of purpose be included on the
+same "printed page" as the copyright notice for easier
+identification within third-party archives.
+
+Copyright {yyyy} {name of copyright owner}
```
kmruiz, line 189: Placeholder placeholderface appeared.

lerouxb:
I'm gonna just close the WIP PRs - I basically had them so I wouldn't lose track of temp branches.

PR #532, opened 2025-05-02
feat: add device ID hashing package COMPASS-8443

gagik:
Creates a consistent, implementation-agnostic hash from a given raw machine ID. The machine ID should originate from `node-machine-id` or `native-machine-id` depending on the platform. The hash is generated using SHA-256 and is designed to be consistent with the Atlas CLI.


packages/device-id/src/get-device-id.ts
```diff
@@ -0,0 +1,95 @@
...
+      onError,
+    }),
+    new Promise<string>((resolve, reject) => {
+      timeoutId = setTimeout(() => {
+        if (onTimeout) {
+          onTimeout(resolve, reject);
+        } else {
+          resolve('unknown');
+        }
+      }, timeout);
```
addaleax, line 34:
```suggestion
      }, timeout).unref?.();
```

addaleax (approved):

packages/device-id/src/get-device-id.ts
gagik: TIL

gagik:
There seems to be some unrelated SBOM dependency failing issue; going ahead with merge.

PR #534, opened 2025-05-05
fix(native-machine-id): replace bindings dependency with explicit paths

Nokel81:
## Description

- Use explicit paths for requiring as that works better with bundlers
- No need to use `bindings` as the alternative is simpler

## Checklist
	- N/A


addaleax (approved):

addaleax:
@Nokel81 Published in `native-machine-id@0.1.2`, thanks for the PR! ✨

Nokel81:
Thanks for the quick response time!

PR #535, opened 2025-05-06
[chore] tweak get-device-id

nirinchev:
## Description
Removes isNodeMachineId arguments and simplifies onTimeout. Also, adds some tests for onTimeout, which were missing.


packages/device-id/src/get-device-id.ts
```diff
@@ -78,10 +68,7 @@ async function resolveMachineId({
 
     // Create a hashed format from the machine ID
     // to match it exactly with the denisbrodbeck/machineid library that Atlas CLI uses.
-    const hmac = createHmac(
-      'sha256',
-      isNodeMachineId ? originalId : originalId,
-    );
+    const hmac = createHmac('sha256', originalId);
```
gagik, line 71: 🙈

packages/device-id/src/get-device-id.ts
```diff
@@ -48,28 +46,20 @@ export function getDeviceId({
...
 async function resolveMachineId({
   getMachineId,
-  isNodeMachineId,
   onError,
 }: GetDeviceIdOptions): Promise<string> {
   try {
-    const originalId = isNodeMachineId
-      ? (await getMachineId())?.toUpperCase()
-      : await getMachineId();
+    const originalId = (await getMachineId())?.toUpperCase();
```
gagik, line 62: I didn't want to do this mainly because I think `native-machine-id` has ability to actually have 1:1 compatibility with the Go project used by Atlas CLI and uppercasing is more a bandaid I'm using on `node-machine-id` which [explicitly applies a lowercase transformation](https://github.com/automation-stack/node-machine-id/blob/f580f9f20668582e9087d92cea2511c972f2e6aa/index.js#L57). 

It's hypothetically possible (though not sure how likely) for what we get from `native-machine-id` (1:1 as Atlas CLI's go plugin) for the device ID to be all lowercase, and we'd be missing out on that otherwise. Though then we'd be incompatible with `node-machine-id` using things...

Best solution here would be to instead merge the relatively small `node-machine-id` code into this and use that instead without the lowercase transformation as a fallback if not native machine ID getter is provided.

packages/device-id/src/get-device-id.ts
```diff
@@ -21,15 +20,14 @@ export function getDeviceId({
...
       onError,
     }),
     new Promise<string>((resolve, reject) => {
       timeoutId = setTimeout(() => {
-        if (onTimeout) {
-          onTimeout(resolve, reject);
-        } else {
-          resolve('unknown');
+        try {
+          resolve(onTimeout?.() ?? 'unknown');
```
gagik, line 28: I don't really think the success of timeout's function should dictate this promise's resolution. If someone tries to write a log and that log write errors, I don't think that says anything about the state of device ID. 

That said, yeah it's weird to outsource the resolution to this function anyhow. so I'd say we should just always resolve to `unknown` for consistency and use onTimeout as a callback without args, we don't this to be too customizable anyhow.

I'd avoid rejecting the promise. I think it's easier to reason about a promise which always resolves than about a promise which can reject. I had problem with rejecting when using it in i.e. mongosh, largely making sure that we can ignore the failed promise as JS does see it as an unresolved async error.
One would need to explicitly handle this rejection with an `await` try/catch or `.catch()`. I think it gets even more complicated with the `Promise.race`, though some solution should be possible.

gagik (approved):

PR #536, opened 2025-05-06
fix: tweak native-machine-id scripts

nirinchev:
## Description
The `check` script was running test, which is inconsistent with the other packages - I've changed it to be consistent with the rest of the packages where it runs typecheck and depcheck, but not test.


packages/native-machine-id/package.json
```diff
@@ -9,11 +9,13 @@
     "bootstrap": "npm run compile",
     "pretest": "npm run compile",
     "install": "node-gyp rebuild",
+    "typecheck": "tsc --noEmit",
     "test": "mocha",
     "test-cov": "nyc -x \"**/*.spec.*\" --reporter=lcov --reporter=text --reporter=html npm run test",
     "test-ci": "npm run test-cov",
     "lint": "eslint . && prettier --check .",
```
addaleax, line 16: We'll also want to add prettier as a dev dependency and add a reformat script, I think

PR #538, opened 2025-05-19
chore: ES2023 in tsconfig and update TS

paula-stacho:
## Description
Looks like the new features are widely supported, except 'WeakMap: Non-registered symbols as keys' in Firefox but that sounds pretty niche?
https://caniuse.com/?search=ES2022
https://caniuse.com/?search=ES2023
https://node.green/

## Checklist


addaleax (approved):

addaleax:
> except 'WeakMap: Non-registered symbols as keys' in Firefox but that sounds pretty niche?

Actually, where are we getting this error? Might be interesting since it could point at a genuine bug (or at least memory leak)

paula-stacho:
> > except 'WeakMap: Non-registered symbols as keys' in Firefox but that sounds pretty niche?
> 
> Actually, where are we getting this error? Might be interesting since it could point at a genuine bug (or at least memory leak)

Ah, we're not getting an error, this was just from the list of features

PR #542, opened 2025-05-30
chore: Deal with there being no known/active connection when doing autocomplete

lerouxb:


packages/mongodb-ts-autocomplete/package.json
```diff
@@ -50,6 +50,7 @@
     "@mongosh/shell-api": "^3.11.0"
   },
   "dependencies": {
+    "@mongosh/shell-api": "^3.11.0",
```
lerouxb, line 53: RE: https://github.com/mongodb-js/devtools-shared/pull/541

Turns out we have a .npmrc file that sets. `legacy-peer-deps=true` for some reason and therefore right now it isn't installing shell-api at all.

lerouxb: Why, though? We import it at runtime.

addaleax: Yeah, as a peer dependency.

If you specify a regular dependency, that's strictly stronger than a peer or dev dependency, so you could just leave it out from the latter categories.

But the goal here is to ensure that our products use a version of the shell-api package that's consistent with what the product uses in other areas of its code; and for that, peer dependencies are the right tool. And if the peer dependency is needed at development time, then peer + dev is the typical pattern for that

PR #546, opened 2025-06-18
fix type exports

lerouxb:


packages/mongodb-ts-autocomplete/package.json
```diff
@@ -25,7 +25,8 @@
   "main": "dist/index.js",
   "exports": {
     "require": "./dist/index.js",
-    "import": "./dist/.esm-wrapper.mjs"
+    "import": "./dist/.esm-wrapper.mjs",
+    "types": "./dist/index.d.ts"
```
lerouxb, line 29: With new ts compiler options:

```
src/index.ts:428:5 - error TS7016: Could not find a declaration file for module '@mongodb-js/mongodb-ts-autocomplete'. '/Users/leroux.bodenstein/mongo/mongosh/packages/autocomplete/node_modules/@mongodb-js/mongodb-ts-autocomplete/dist/.esm-wrapper.mjs' implicitly has an 'any' type.
  There are types at '/Users/leroux.bodenstein/mongo/mongosh/packages/autocomplete/node_modules/@mongodb-js/mongodb-ts-autocomplete/dist/index.d.ts', but this result could not be resolved when respecting package.json "exports". The '@mongodb-js/mongodb-ts-autocomplete' library may need to update its package.json or typings.
```

I just changed the others where we had types at the top but not inside exports too. It is a bit of a mixed bag across the packages.

packages/dl-center/package.json
```diff
@@ -25,7 +25,8 @@
   "main": "dist/index.js",
   "exports": {
     "require": "./dist/index.js",
-    "import": "./dist/.esm-wrapper.mjs"
+    "import": "./dist/.esm-wrapper.mjs",
+    "types": "./dist/index.d.ts"
```
lerouxb, line 29: Depending on the module resolution settings you pass to the ts compiler settings it might fall back to the types at the root level or it might not. With node16 rules it now _tells_ you about the types at the root level, but doesn't use it. 🙄 I don't think it affects every package because we don't use them all in mongosh, but I figured for consistency I'm just specifying the types everywhere we specify exports. We already did it in a bunch of packages anyway.

PR #547, opened 2025-06-18
shell-api 3.13.0

lerouxb:
mongodb-ts-autocomplete now doesn't want to install from shell-api because the peer dep version doesn't match. I guess this will then break again the moment shell-api bumps after merging the PR there..

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8104,9 +8104,9 @@
       }
     },
     "node_modules/@mongosh/shell-api": {
-      "version": "3.11.0",
-      "resolved": "https://registry.npmjs.org/@mongosh/shell-api/-/shell-api-3.11.0.tgz",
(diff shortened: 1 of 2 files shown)


addaleax:
> mongodb-ts-autocomplete now doesn't want to install from shell-api because the peer dep version doesn't match

Err, it might be worth digging into this deeper ... this should not be happening 🙂 

Aside, we're in an awkward situation here because I messed up the last release and 3.13 isn't actually the latest release, 3.12 is ([Slack thread](https://mongodb.slack.com/archives/GJ2CBBY86/p1749813643486269)). And that should get resolved by the next mongosh release, like, almost immediately, but that is blocked on Evergreen being buggy and not picking up the release ([Slack thread](https://mongodb.slack.com/archives/C0V896UV8/p1750245085548739))

addaleax (approved):

PR #549, opened 2025-06-19
fix exports in package.json to work with ts's implementation of node16 moduleResolution

lerouxb:
This is a follow-up to https://github.com/mongodb-js/devtools-shared/pull/546 where I didn't properly understand this.

The context of why/where I need this is to be able to use a separate types export from the default one [from mongosh](https://github.com/mongodb-js/mongosh/pull/2478/files). 

The latest error I'm getting there:

```
src/index.ts:428:5 - error TS7016: Could not find a declaration file for module '@mongodb-js/mongodb-ts-autocomplete'. '/Users/leroux.bodenstein/mongo/mongosh/packages/autocomplete/node_modules/@mongodb-js/mongodb-ts-autocomplete/dist/.esm-wrapper.mjs' implicitly has an 'any' type.
  There are types at '/Users/leroux.bodenstein/mongo/mongosh/packages/autocomplete/node_modules/@mongodb-js/mongodb-ts-autocomplete/dist/index.d.ts', but this result could not be resolved when respecting package.json "exports". The '@mongodb-js/mongodb-ts-autocomplete' library may need to update its package.json or typings.

428     '@mongodb-js/mongodb-ts-autocomplete'
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Basically TS _sees_ the right types but refuses to use it 😛 

The way this should work is actually, if you're exporting multiple things:

```javascript
exports: {
  '.': {
    import: {
      default: '',
      types: '',
    },
    require: {
      default: '',
      types: '',
    }
  }
  foo: {
    import: {
      default: '',
      types: '',
    },
    require: {
      default: '',
      types: '',
    }
  }
}
```

then if you're just exporting one thing (the usual case):

```javascript
exports: {
  import: {
    default: '',
    types: '',
  },
  require: {
    default: '',
    types: '',
  }
}
```

Then there's also "main" and "types" for things that don't support "exports" yet.

References:
* https://www.typescriptlang.org/tsconfig/#moduleResolution (not much there, unfortunately)
* https://github.com/microsoft/TypeScript/issues/52363#issuecomment-1659179354 and the subsequent comment.
* https://github.com/addaleax/gen-esm-wrapper

### DIFF
--- a/packages/device-id/package.json
+++ b/packages/device-id/package.json
@@ -23,14 +23,19 @@
   ],
   "license": "Apache-2.0",
   "main": "dist/index.js",
+  "types": "./dist/index.d.ts",
   "exports": {
     ".": {
-      "require": "./dist/index.js",
-      "import": "./dist/.esm-wrapper.mjs",
-      "types": "./dist/index.d.ts"
+      "require": {
+        "default": "./dist/index.js",
+        "types": "./dist/index.d.ts"
+      },
+      "import": {
+        "default": "./dist/.esm-wrapper.mjs",
+        "types": "./dist/index.d.ts"
+      }
     }
   },
-  "types": "./dist/index.d.ts",
   "scripts": {
     "bootstrap": "npm run compile",
     "prepublishOnly": "npm run compile",
--- a/packages/devtools-proxy-support/package.json
+++ b/packages/devtools-proxy-support/package.json
@@ -23,19 +23,29 @@
   ],
   "license": "Apache-2.0",
   "main": "dist/index.js",
+  "types": "./dist/index.d.ts",
   "exports": {
     ".": {
-      "require": "./dist/index.js",
-      "import": "./dist/.esm-wrapper.mjs",
-      "types": "./dist/index.d.ts"
+      "require": {
+        "default": "./dist/index.js",
+        "types": "./dist/index.d.ts"
+      },
+      "import": {
+        "default": "./dist/.esm-wrapper.mjs",
+        "types": "./dist/index.d.ts"
+      }
     },
     "./proxy-options": {
-      "require": "./dist/proxy-options-public.js",
-      "import": "./dist/.esm-wrapper-po.mjs",
-      "types": "./dist/proxy-options-public.d.ts"
+      "require": {
+        "default": "./dist/proxy-options-public.js",
+        "types": "./dist/proxy-options-public.d.ts"
+      },
+      "import": {
+        "default": "./dist/.esm-wrapper-po.mjs",
+        "types": "./dist/proxy-options-public.d.ts"
+      }
     }
   },
-  "types": "./dist/index.d.ts",
   "scripts": {
     "bootstrap": "npm run compile",
     "prepublishOnly": "npm run compile",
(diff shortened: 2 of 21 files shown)


packages/device-id/package.json
```diff
@@ -23,14 +23,19 @@
...
   "main": "dist/index.js",
+  "types": "./dist/index.d.ts",
   "exports": {
     ".": {
-      "require": "./dist/index.js",
-      "import": "./dist/.esm-wrapper.mjs",
-      "types": "./dist/index.d.ts"
+      "require": {
+        "default": "./dist/index.js",
+        "types": "./dist/index.d.ts"
```
lerouxb, line 31: I don't think anything that's going to use exports.require is actually going to pick up this types field because the default / "node" moduleResolution in typescript was clearly happily using "types" from the root and now with node16 it will use import.types. plus node/js (outside of typescript) is never going to import the types anyway but it is here for consistency.

packages/device-id/package.json
```diff
@@ -23,14 +23,19 @@
...
+      },
+      "import": {
+        "default": "./dist/.esm-wrapper.mjs",
+        "types": "./dist/index.d.ts"
```
lerouxb, line 35: This comment [here](https://github.com/microsoft/TypeScript/issues/52363#issuecomment-1659227177) says that you need separate types files for require/import but so far in my testing by just npm linking things together it seems to work?

What I'll do once this PR lands is to make sure we bump all these packages where they are used so that I find and fix any errors and not someone else.

packages/device-id/package.json
```diff
@@ -23,14 +23,19 @@
...
+  "types": "./dist/index.d.ts",
   "exports": {
     ".": {
-      "require": "./dist/index.js",
-      "import": "./dist/.esm-wrapper.mjs",
-      "types": "./dist/index.d.ts"
+      "require": {
+        "default": "./dist/index.js",
+        "types": "./dist/index.d.ts"
+      },
```
kraenhansen, line 32: A "types" condition like this is redundant, since TSC will automatically infer it from the .js file.

```suggestion
      "require": "./dist/index.js",
```

addaleax (approved):

gribnoysup (approved):

packages/device-id/package.json
lerouxb: so just for the "require" ones? I left a comment elsewhere where I already had a suspicion like that.

packages/device-id/package.json
gagik: > A "types" condition like this is redundant, since TSC will automatically infer it from the .js file.

is this the case? I have had issues with packages which don't explicitly define it before

gagik (approved):

lerouxb:
I'm gonna merge this because I'm kinda blocked on it right now, but I'll keep speaking to @kraenhansen about his feedback and come around to it.

packages/device-id/package.json
kraenhansen: Chatted with @lerouxb offline, I'd be happy to share an example for anyone interested.

PR #554, opened 2025-06-30
fix: explicitly specify and bump typescript related deps, fix debug filtering

lerouxb:


packages/ts-autocomplete/package.json
```diff
@@ -54,7 +54,7 @@
   "dependencies": {
     "debug": "^4.4.0",
     "lodash": "^4.17.21",
-    "typescript": "^5.0.4"
+    "typescript": "^5.3.3"
```
lerouxb, line 57: Same version as in mongosh.

packages/ts-autocomplete/src/index.ts
```diff
@@ -165,10 +165,7 @@ function filterDiagnostics(diagnostics: ts.Diagnostic[]): {
       ..._.pick(item, 'messageText'),
     };
 
-    if (
-      result.fileName === '/shell-api.ts' ||
-      result.fileName?.startsWith('/connection-')
-    ) {
+    if (result.fileName?.endsWith('shell-api.ts')) {
```
lerouxb, line 168: connection files actually start with connection_, not connection- but it is only the -shell-api.ts ones that are huge and clog up debug output, causing it to be cut off and then we can't see all the actual errors.

packages/ts-autocomplete/package.json
addaleax: Here in the repo the package-lock points to 5.8.2 anyway, feel free to also bump it to that in mongosh too (not a blocker for this PR really, the version here in package.json doesn't really matter anyway)

addaleax (approved):

PR #555, opened 2025-06-30
chore: figure out all the types we need at runtime, bundle them as giant strings, serve them to the language server MONGOSH-2350

lerouxb:
The typescript compiler implicitly loads the files for the javascript library (that gives you things like Date, Regex, Promise, etc.) and the node types. shell-api and mongodb uses the node-types and mongodb. mql also uses mongodb which in turn uses node types. Shell-api also uses things like Promise.

(Aside: Not all of these will be used by editors in Compass, most only apply to shell autocompleter.)

In the built/bundled mongosh binary we can't have the typescript language server read files because those files won't exist on a user's system. So we have to make sure it loads the files from our in-memory "code holder".

Which means we have to read these files and store them in a module as giant strings so that they will be included in the binary. In that sense it is analogous to what we're already doing for bson, mql, mongodb and shell-api.

### DIFF
--- a/packages/ts-autocomplete/src/index.ts
+++ b/packages/ts-autocomplete/src/index.ts
@@ -114,22 +114,13 @@ function getVirtualLanguageService(
       }
     },
     readDirectory: (...args) => {
-      if (fallbackServiceHost && fallbackServiceHost.readDirectory) {
-        return fallbackServiceHost.readDirectory(...args);
-      }
-      return [];
+      return fallbackServiceHost?.readDirectory?.(...args) ?? [];
     },
     directoryExists: (...args) => {
-      if (fallbackServiceHost && fallbackServiceHost.directoryExists) {
-        return fallbackServiceHost.directoryExists(...args);
-      }
-      return false;
+      return fallbackServiceHost?.directoryExists?.(...args) ?? false;
     },
     getDirectories: (...args) => {
-      if (fallbackServiceHost && fallbackServiceHost.getDirectories) {
-        return fallbackServiceHost.getDirectories(...args);
-      }
-      return [];
+      return fallbackServiceHost?.getDirectories?.(...args) ?? [];
     },
     log: (...args) => debugLog(args),
     trace: (...args) => debugTrace(args),


packages/ts-autocomplete/src/index.ts
```diff
@@ -10,20 +11,52 @@ type TypeFilename = string;
...
+// TODO
+/*
+function createIOError(code: string, details = ""): NodeJS.ErrnoException {
+    const err: NodeJS.ErrnoException = new Error(`${code} ${details}`);
+    err.code = code;
+    if (Error.captureStackTrace) Error.captureStackTrace(err, createIOError);
+    return err;
+}
+    */
+function relativeNodePath(fileName: string): string {
```
lerouxb, line 23: This whole function is kinda the crux of this. If we can't reliably map what ts is trying to read to something we have in our fixture of types both when testing and when doing npm run start in mongosh AND in the bundled mongosh, then none of this will work.

lerouxb: To test this I copied these packages to mongosh and built a mongosh binary there. It seemed to work. For a full test we'll have to do a release of this then bump it in mongosh.

packages/ts-autocomplete/src/index.ts
```diff
@@ -8,66 +9,144 @@ const debugError = createDebug('ts-autocomplete:error');
...
     },
     getScriptSnapshot: (fileName) => {
+      fileName = relativeNodePath(fileName);
       if (fileName in codeHolder) {
-        return ts.ScriptSnapshot.fromString(codeHolder[fileName]);
+        return ts.ScriptSnapshot.fromString(codeHolder[fileName].toString());
       }
 
-      return ts.ScriptSnapshot.fromString(ts.sys.readFile(fileName) || '');
+      if (process.env.CI) {
```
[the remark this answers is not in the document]
lerouxb: I guess the obvious solution is to pass in an optional fallback "resolver" from the tests. So it would be an option on the class.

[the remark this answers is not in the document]
gribnoysup: Yeah, I agree that allowing to pass through a custom servicesHost and doing it from tests would be better than relying on env

packages/ts-autocomplete/src/index.ts
```diff
@@ -8,66 +9,145 @@ const debugError = createDebug('ts-autocomplete:error');
...
+        const result = ts.ScriptSnapshot.fromString(
+          // NOTE: some files do not exist. A good example is "typescript/lib/es2023.ts"
```
gribnoysup, line 92: Just FYI, maybe you already know this, I think when you see stuff like just `es2023` here it means that you need to resolve it to one of those lists of other files that we're currently listing manually, I'm pretty sure there's a method for that on ts or ts.sys that you can use, but not probably something we need to spend time figuring out right now, idk

packages/ts-autocomplete/src/index.ts
```diff
@@ -8,66 +9,145 @@ const debugError = createDebug('ts-autocomplete:error');
...
+        return result;
+      }
+
+      return false;
     },
     readFile: (fileName) => {
+      fileName = relativeNodePath(fileName);
       if (fileName in codeHolder) {
-        return codeHolder[fileName];
+        return codeHolder[fileName].toString();
```
gribnoysup, line 124: I'm guessing this is because we now also have these boolean values there for exists case? Would it make sense to throw if you're trying to readFile that we didn't read actually or is it an expected thing to return stringified boolean here? (I'm maybe just not understanding this part fully)

packages/ts-autocomplete/src/index.ts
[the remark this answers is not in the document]
lerouxb: done now

packages/mongodb-ts-autocomplete/scripts/extract-types.ts
```diff
@@ -1,43 +1,257 @@
...
+  ],
+  'undici-types': ['package.json', 'index.d.ts'],
+  url: [
+    'package.json',
+    'url.js', // exists only
+  ],
+  util: [
+    'package.json',
+    'util.js', // exists only
+  ],
```
addaleax, line 214: I'm sure you've had a similar thought, but is there something we can do to make this list a bit easier to maintain (i.e. auto-generate it)?

packages/mongodb-ts-autocomplete/scripts/extract-types.ts
```diff
@@ -1,43 +1,257 @@
...
+    'wasi.d.ts',
+    'worker_threads.d.ts',
+    'zlib.d.ts',
+  ],
+  assert: [
+    'package.json',
+    'assert.js', // exists only
+  ],
+  buffer: ['package.json', 'index.d.ts'],
+  events: ['package.json'],
```
gribnoysup, line 144: Not sure how important this is, probably not much, but I think these should be resolving to node built-ins and not to the npm packaged reimplementations of those, just wondering why they are getting here

gribnoysup (approved):
Looks good to me as much as I understand how all these things are wired together. Anna has a good point about trying to autogenerate the files list, maybe we can grep those dependency dirs for all js, json, and d.ts files to build this list or something like that

packages/mongodb-ts-autocomplete/scripts/extract-types.ts
lerouxb: Best idea I've had so far (and I thought about this for a couple of days, trying different things) is the fallback servicesHost that gathers the ones we're missing and fails a test. That's exactly how I build this list.

I don't expect this to break at all unless we bump typescript or @types/node and even then probably rarely. Personally I think we can see how it goes and optimise that at a later stage if it becomes a problem.

Whatever we do I'd like to keep a human in the loop because automatically shipping everything typescript decides to touch during module resolution is probably at least a little bit dangerous.

lerouxb: But I'm definitely open to suggestions and happy to build whatever people come up with!

lerouxb:
This is green and approved so I think I'll merge so that I can test it out in mongosh to make 100% sure this will even work. We can keep refining the way we decide what needs bundling.

PR #557, opened 2025-07-02
fix(devtools-connect): skip driver ping during connect COMPASS-9536

nbbeeken:
## Description

https://github.com/mongodb-js/compass/pull/7083

Set the option that makes the driver _not_ run ping on connect.

This removes some errors that could happen:
- Ping tests for authentication because it is a command that runs through the connection pool the connection created must be auth-ed
- The server selected is the primary unless a client-wide read preference was set. This blocks the ping from running until a primary is discovered

This improves some scenarios:
- No longer is a pool (likely the primary) pre populated with one connection when connect returns
- Connect returns sooner than RTT to any particular node since it does not wait on monitoring / server selection 
  - Nor does connect wait for a MongoDB handshake RTT in addtion to the ping RTT

## Open Questions

Does the error timing matter to downstream projects? 

Any command can fail auth at anytime, and a primary may not be discoverable at boot but could be when you issue a command. Is "connect time" erroring important? 

## Checklist

### DIFF
--- a/packages/devtools-connect/src/connect.spec.ts
+++ b/packages/devtools-connect/src/connect.spec.ts
@@ -51,7 +51,12 @@ describe('devtools connect', function () {
       expect(mClientType.getCalls()[0].args[0]).to.equal(uri);
       expect(
         Object.keys(mClientType.getCalls()[0].args[1]).sort(),
-      ).to.deep.equal(['allowPartialTrustChain', 'ca', 'lookup']);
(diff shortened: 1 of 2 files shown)


addaleax (approved):
This LGTM in the sense that I think we can drop one of these pings, but we probably want to finish the conversation in https://github.com/mongodb-js/compass/pull/7083/files before merging either of these PRs

PR #559, opened 2025-07-04
chore: test that autocomplete works without the fallback, re-order the tests

lerouxb:
Just a follow-up to make sure that we test this properly.

### DIFF
--- a/packages/mongodb-ts-autocomplete/scripts/extract-types.ts
+++ b/packages/mongodb-ts-autocomplete/scripts/extract-types.ts
@@ -141,7 +141,10 @@ const deps: Record<string, string[]> = {
     'assert.js', // exists only
   ],
   buffer: ['package.json', 'index.d.ts'],
-  events: ['package.json'],
+  events: [
+    'package.json',
+    'events.js', // exists only (also only on windows)
+  ],
   punycode: [
     'package.json',
     'punycode.js', // exists only


packages/mongodb-ts-autocomplete/src/autocompleter.spec.ts
```diff
@@ -126,11 +128,21 @@ describe('MongoDBAutocompleter', function () {
...
+    });
+
+    autocompleterWithFallback = new MongoDBAutocompleter({
+      context: autocompleterContext,
       fallbackServiceHost,
     });
   });
 
-  afterEach(function () {
+  it('autocompletes', async function () {
```
lerouxb, line 139: The best way to fully test/prove this automatically (ie. that it can recreate the list) is do comment out deps in extract-types.ts to make it just an empty structure, run extract-types so none of the @types/node and lib files are in there, then run this test. It then finds the entire list.

I tried modifying extract-types to calculate the list manually by running autocomplete, but it becomes a bit code-surgery with things passed down just for testing and a chicken and egg problem of extract-types using autocomplete which imports the extracted types which aren't generated yet... I tried to fix it with dynamic import, but then TS doesn't want to import the code from the script and it just becomes a rabbit hole.

I think this test that lists what's missing even if you start over with an empty list and the rest of the tests fail because they don't use the fallback is probably good enough.

PR #562, opened 2025-07-07
fix: when autocompleting with insertText, use that rather than name

lerouxb:
`db.fo` -> `db["foo.bar"]` rather than `db.foo.bar` so that the result is valid and then if you type on like `db["foo.bar"].fi` it will continue to autocomplete as `db["foo.bar"].find`.

Because `db.foo.bar.fi` won't autocomplete because `db.foo.bar` is invalid if the collection name is `foo.bar`. ie. there is no prop `foo` on `db`.

### DIFF
--- a/packages/ts-autocomplete/src/index.spec.ts
+++ b/packages/ts-autocomplete/src/index.spec.ts
@@ -83,6 +83,11 @@ describe('Autocompleter', function () {
       expect(autoCompleter.autocomplete('myGlobalObject.')).to.deep.equal([
         {
           kind: 'property',
+          name: 'foo.bar',
+          result: 'myGlobalObject["foo.bar"]',
+        },
+        {
+          kind: 'property',
           name: 'functionProp',
           result: 'myGlobalObject.functionProp',
         },
@@ -100,6 +105,11 @@ describe('Autocompleter', function () {
       });
 
       expect(autoCompleter.autocomplete('myGlobalObject.str')).to.deep.equal([
+        {
+          kind: 'property',
+          name: 'foo.bar',
+          result: 'myGlobalObject["foo.bar"]',
+        },
         {
           kind: 'property',
           name: 'functionProp',
@@ -121,6 +131,11 @@ describe('Autocompleter', function () {
       expect(
         autoCompleter.autocomplete('myGlobalObject.doesNotExist'),
       ).to.deep.equal([
+        {
+          kind: 'property',
+          name: 'foo.bar',
+          result: 'myGlobalObject["foo.bar"]',
+        },
         {
           kind: 'property',
           name: 'functionProp',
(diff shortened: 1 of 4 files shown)


packages/ts-autocomplete/src/index.spec.ts
```diff
@@ -207,6 +222,32 @@ describe('Autocompleter', function () {
...
+
+      expect(completions).to.deep.equal([
+        {
+          kind: 'property',
+          name: 'foo.bar',
+          result: 'myGlobalObject["foo.bar"]',
+        },
+        {
+          kind: 'property',
+          name: 'functionProp',
```
nirinchev, line 241: Hm... why are we getting these completions? Shouldn't they be filtered out since they don't start with `fo`?

packages/ts-autocomplete/src/index.ts
```diff
@@ -190,8 +190,23 @@ function mapCompletions(
   return completions.entries
     .filter((entry) => filter({ trigger, kind: entry.kind, name: entry.name }))
     .map((entry) => {
+      /*
+      With the includeCompletionsWithInsertText option set, auto-completions
+      might have an insertText property for cases like where a property has
```
nirinchev, line 195: What other cases are there and should we add tests for those as well?

nirinchev (approved):
Looks good - just non-blocking comments on my part.

packages/ts-autocomplete/src/index.spec.ts
lerouxb:
```
function filterStartingWith({
  name,
  trigger,
}: {
  name: string;
  trigger: string;
}): boolean {
  name = name.toLocaleLowerCase();
  trigger = trigger.toLocaleLowerCase();

  if (name.startsWith(trigger)) {
    console.log({ name, trigger });
    return true;
  };
  return false;
}
```

trigger is a blank string in this case:

```
  Autocompleter
    without filter
      ✓ returns the global scope for a global variable that does not exist
      ✓ returns completions for global variables
      ✓ returns nothing for a member of a variable that does not exist
      ✓ returns matches for a member of a variable that exists
      ✓ returns matches for part of a member of a variable that exists
      ✓ returns matches for an unknown member of a variable that exists
      ✓ returns the global scope for object parameters of a function that does not exist
      ✓ returns matches for object parameters of a function that exists
      ✓ returns matches for part of an object parameter of a function that exists
      ✓ returns matches for an unknown object parameter of a function that exists
      ✓ returns completions for variables with dots in their names
    with filter
      ✓ returns nothing when it has no code
{ name: 'myglobalfunction', trigger: 'myglobalfunct' }
      ✓ returns completions for global variables
{ name: 'myglobalobject', trigger: 'myglobalobject' }
{ name: 'foo.bar', trigger: '' }
{ name: 'functionprop', trigger: '' }
{ name: 'stringprop', trigger: '' }
{ name: 'functionprop', trigger: 'functionpr' }
      ✓ returns completions for members of global variables
{ name: 'param1', trigger: 'par' }
{ name: 'param2', trigger: 'par' }
      ✓ returns completions for the fields of object function parameters


  15 passing (60ms)
```

packages/ts-autocomplete/src/index.ts
lerouxb: probably dashes, spaces, quotes..

nirinchev: Actually, will be interesting to see some corner cases involving quotes - e.g. a collection called `foo"bar'` fails with the current autocompleter - kind of hoping it will work out of the box with the new one

PR #565, opened 2025-08-05
feat(mongodb-log-writer): add support for recursively cleaning up old log files MCP-103

nirinchev:
## Description
This allows the log manager to optionally consider log files inside child directories. This is related to MCP-103 where we will be creating child log managers writing files inside subdirectories for specific sessions, while the parent log manager will be responsible for cleaning up all logs.

## Checklist

### DIFF
--- a/packages/mongodb-log-writer/src/mongo-log-manager.ts
+++ b/packages/mongodb-log-writer/src/mongo-log-manager.ts
@@ -47,14 +47,30 @@ export class MongoLogManager {
     this._options = options;
   }
 
-  private async deleteFile(path: string): Promise<void> {
+  private async deleteFile(filePath: string): Promise<void> {
     try {
-      await fs.unlink(path);
+      await fs.unlink(filePath);
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch (err: any) {
       if (err?.code !== 'ENOENT') {
-        this._options.onerror(err as Error, path);
+        this._options.onerror(err as Error, filePath);
       }
+    }
+
+    let basePath = path.dirname(filePath);
+    while (basePath !== this._options.directory) {
+      // Delete empty directories up to the base directory
+      try {
+        await fs.rmdir(path.dirname(filePath));
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } catch (err: any) {
+        // If the directory is not empty, we stop deleting.
+        if (err?.code === 'ENOTEMPTY' || err?.code === 'ENOENT') {
+          break;
+        }
+      }
+
+      basePath = path.dirname(basePath);
     }
   }
 
--- a/packages/mongodb-log-writer/src/mongo-log-manager.ts
+++ b/packages/mongodb-log-writer/src/mongo-log-manager.ts
@@ -1,6 +1,7 @@
 import path from 'path';
 import { ObjectId } from 'bson';
 import { once } from 'events';
+import type { Dir } from 'fs';
 import { createWriteStream, promises as fs } from 'fs';
 import { createGzip, constants as zlibConstants } from 'zlib';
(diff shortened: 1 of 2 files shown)


packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -39,42 +40,63 @@ export class MongoLogManager {
...
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch (err: any) {
       if (err?.code !== 'ENOENT') {
-        this._options.onerror(err as Error, path);
+        this._options.onerror(err as Error, filePath);
       }
     }
+
+    let basePath = path.dirname(filePath);
+    while (basePath !== this._options.directory) {
```
addaleax, line 61: What if `this._options.directory` was specified as `/foo/bar/`? Repeatedly calling `path.dirname()` on `/foo/bar/baz` will never reach that filename.

I'd imagine that `path.normalize()` together with some sort of prefix check would work here, but I'm not sure if that's the most reliable way to do this either

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -39,42 +40,63 @@ export class MongoLogManager {
...
+      // Delete empty directories up to the base directory
+      try {
+        await fs.rmdir(path.dirname(filePath));
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } catch (err: any) {
+        // If the directory is not empty, we stop deleting.
+        if (err?.code === 'ENOTEMPTY' || err?.code === 'ENOENT') {
```
addaleax, line 68: These would be the most common errors that `rmdir()` could encounter, but I don't think it makes sense to continue in _any_ error situation here, right?

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -39,42 +40,63 @@ export class MongoLogManager {
...
     // Delete files older than N days
     const deletionCutoffTimestamp =
       deletionStartTimestamp - this._options.retentionDays * 86400 * 1000;
 
     const dir = this._options.directory;
-    let dirHandle;
+    let dirHandle: Dir;
     try {
-      dirHandle = await fs.opendir(dir);
+      dirHandle = await fs.opendir(dir, { recursive });
```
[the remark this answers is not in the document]
nirinchev: Yeah, that's also one of my concerns. I'll try and explore a different approach here - might be a better way to handle things, so will put that back in draft.

packages/mongodb-log-writer/src/mongo-log-manager.ts
```diff
@@ -101,18 +123,25 @@ export class MongoLogManager {
...
         // where lots and lots of log files end up and filesystem operations happen
         // with network latency.
-        if (Date.now() - deletionStartTimestamp > maxDurationMs) break;
+        if (Date.now() - deletionStartTimestamp > maxDurationMs) {
+          break;
+        }
+
+        if (!dirent.isFile()) {
+          continue;
+        }
```
addaleax, line 132: This leaves us in a situation where, if removing a directory fails for any reason, we're never going to clean that directory up. I think that may be something we just need to accept, unless we want to start e.g. imposing naming restrictions on the nested subdirectories – that's something we should probably decide on now, though.

nirinchev:
Moving that back into draft - I feel I may have jumped the gun on that approach and I'm not super happy with some of the implications from this change. I'll think a bit more about it and see if I can come up with something better.

nirinchev:
Closing this as I've ultimately decided to pursue the other approach, where we log from all sessions in the same log file, but we'll be attributing log messages from different sessions with their ids.

PR #572, opened 2025-09-17
feat(mongodb-ns)!: move ns package

johnjackweir:
Move the https://github.com/mongodb-js/ns package into here where it belongs

Major version bump for license change


packages/ns/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "mongodb.js"
+  ],
+  "license": "MIT",
+  "dependencies": {},
+  "devDependencies": {
+    "eslint-config-mongodb-js": "^5.0.3",
+    "mocha": "^7.0.0",
+    "mongodb-js-precommit": "^2.0.0"
+  }
+}
```
addaleax, line 31: This works, but honestly, I'd recommend just using the create-workspace script to create this directory and then copy the README + index.js + index.d.ts + test.js into the created directory. That's what I've done in the past and it's not too much effort but is overall much nicer because then you end up with a package that's aligned with our general workflow, coding style, etc.

Optional but I'd definitely consider it

johnjackweir: ah that's what we meant by workspace! let me try that now

addaleax: Yeah, we're adopting the terminology of npm and similar tools here: https://docs.npmjs.com/cli/v8/using-npm/workspaces

addaleax (approved):

nbbeeken (approved):

packages/mongodb-ns/src/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  validDatabaseName: boolean;
+  validCollectionName: boolean;
+  databaseHash: number;
+  toString(): string;
+  // Assigned in the constructor, but will always be undefined
+  isConf(): undefined;
+};
+
+type NSConstructor = {
+  new (ns: string | NS): NS;
```
nbbeeken, line 26:
```suggestion
type NSConstructor = {
  (ns: string | NS): NS;
  new (ns: string | NS): NS;
```

Sorry!! one more thing, since our goal is to NOT change all the call sites of this.

since compass almost universally imports like:
```ts
import toNS from 'mongodb-ns';
// and calls it like
toNS('string');
```

If we don't add the plain `(ns: string | NS): NS;` in addition to the `new` one you'll be required to add `new` everywhere by TS.

nbbeeken (changes requested):

PR #574, opened 2025-09-23
feat(mongodb-runner): add downloadDir option MONGOSH-1875

gagik:
There are cases where the place we want to manage the place to download and store binaries ourselves. For example in mongosh this is useful in making sure each `mongodb-runner` has its own temporary directory for all other files but refers to only 1 download directory for the database binaries.

### DIFF
--- a/packages/mongodb-runner/src/mongocluster.ts
+++ b/packages/mongodb-runner/src/mongocluster.ts
@@ -93,7 +93,7 @@ export class MongoCluster {
     const cluster = new MongoCluster();
     cluster.topology = options.topology;
     if (!options.binDir) {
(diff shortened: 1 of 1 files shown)


packages/mongodb-runner/src/mongocluster.ts
```diff
@@ -84,8 +93,8 @@ export class MongoCluster {
     const cluster = new MongoCluster();
     cluster.topology = options.topology;
     if (!options.binDir) {
-      options.binDir = await downloadMongoDb(
-        options.tmpDir,
+      options.binDir = await MongoCluster.downloadMongoDb(
```
addaleax, line 96:
```suggestion
      options.binDir = await this.downloadMongoDb(
```

PR #633, opened 2026-04-14
feat: add script to check transitive deps COMPASS-10569

lerouxb:
COMPASS-10569

Here it is on Compass: https://gist.github.com/lerouxb/d076c160f49cc1973db0490cab2bcf2a

We probably want to run it with `--ignore-dev-deps`. Without it is a convenient way to get errors during testing, though.


.check-transitive-deps.json
```diff
@@ -0,0 +1,9 @@
+{
+  "deps": [
```
lerouxb, line 2: Anything else we want to keep aligned? Running this with just `["*"]` is kinda hilarious.

packages/monorepo-tools/src/utils/get-packages-in-topological-order.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+    const pkgDir = path.join(root, dir);
+    await fs.mkdir(pkgDir, { recursive: true });
+    await fs.writeFile(
+      path.join(pkgDir, 'package.json'),
+      JSON.stringify(packageJson),
+    );
+  }
+}
+
+describe('getPackagesInTopologicalOrder', function () {
```
lerouxb, line 28: I added these tests as a drive-by because coveralls was complaining about it and that coverage is down. Then I convinced myself that I started importing this from tests which made it reachable and counted towards coverage... But I think that code's gone if it was ever the case.

Anyway. Here are some tests for it of dubious quality.

gribnoysup (approved):
Can't say I followed the whole logic very thoroughly, but overall looks LGTM. I think some helper logic here already exists in semver-helpers or depalign scripts in similar form, but slightly different implementation, but I don't have a very strong preference for trying to align and share it all

addaleax (approved):

PR #711, opened 2026-04-28
feat(constants): Auto embedding public preview templates COMPASS-10333

mongodbnolanc:
## Description
Adds two new constants, both intended to be used for auto embedding public preview features in Compas.
- Adds the auto embedding version of the vector search index template
- Adds an alternate version of $vectorSearch stage which includes

## Checklist

- [x] Updated existing tests


packages/mongodb-constants/src/stage-operators.ts
```diff
@@ -1296,6 +1296,46 @@ const STAGE_OPERATORS = [
...
+  env: [ATLAS, ON_PREM],
+  meta: 'stage',
+  version: '>=6.0.10 <7.0.0 || >=7.0.2',
+  apiVersions: [],
+  namespaces: [COLLECTION, VIEW],
+  description: 'Performs a kNN search on embeddings in the specified field(s)',
+  comment: `/**
+/**
+ * query: {“text”: Natural Language Query string for performing semantic search} Use this option if using Automated Embedding. 
+ * queryVector: Array of numbers of BSON types \`int\` or \`double\` that represent the query vector. The array size must match the number of vector dimensions specified in the index for the field. Use this option if you're bringing your own vectors.
```
johnjackweir, line 1316: should this also be marked as required?

packages/mongodb-constants/src/stage-operators.ts
```diff
@@ -1296,6 +1296,46 @@ const STAGE_OPERATORS = [
...
+  score: 1,
+  env: [ATLAS, ON_PREM],
+  meta: 'stage',
+  version: '>=6.0.10 <7.0.0 || >=7.0.2',
+  apiVersions: [],
+  namespaces: [COLLECTION, VIEW],
+  description: 'Performs a kNN search on embeddings in the specified field(s)',
+  comment: `/**
+/**
+ * query: {“text”: Natural Language Query string for performing semantic search} Use this option if using Automated Embedding.
```
johnjackweir, line 1315: I don't see this option mentioned in docs, should those be updated?

packages/mongodb-constants/src/stage-operators.ts
```diff
@@ -1296,6 +1296,46 @@ const STAGE_OPERATORS = [
...
+ * index: Name of the Atlas Search index to use. (Required)
+ * limit: Number (of type \`int\` only) of documents to return in the results. (Required)
+ * model: Specify a model, compatible with the one used during index creation, to perform embedding generation of the \`query\` string. If nothing is specified the same model used during index creation is used. Available only if using Automated Embedding. 
+i * filter: Any MongoDB Query Language (MQL) match expression that compares an indexed field with a boolean, number (not decimals), or string to use as a prefilter. (Optional)
+ * exact: Choose between false for ANN (Approximate Nearest Neighbor) and true for ENN (Exact Nearest Neighbor). Defaults to false. (Optional)
+ */
+`,
+  snippet: `{
+  //query: {"text": \${1:query string}},
+  //queryVector: [\${2:dimension1}, \${3:dimension2}, ...],
```
Anemy: Most of the time slash quotes will have a space before the content, how about:
```suggestion
  // query: {"text": \${1:query string}},
  // queryVector: [\${2:dimension1}, \${3:dimension2}, ...],
```

packages/mongodb-constants/src/stage-operators.ts
```diff
@@ -1296,6 +1296,46 @@ const STAGE_OPERATORS = [
...
+  comment: `/**
+/**
+ * query: {“text”: Natural Language Query string for performing semantic search} Use this option if using Automated Embedding. 
+ * queryVector: Array of numbers of BSON types \`int\` or \`double\` that represent the query vector. The array size must match the number of vector dimensions specified in the index for the field. Use this option if you're bringing your own vectors.
+
+ * path: The field to search. (Required)
+ * numCandidates: Number of nearest neighbors to use during the search. You can specify a number higher than the number of documents to return (\`limit\`) to increase accuracy. (Required)
+ * index: Name of the Atlas Search index to use. (Required)
+ * limit: Number (of type \`int\` only) of documents to return in the results. (Required)
+ * model: Specify a model, compatible with the one used during index creation, to perform embedding generation of the \`query\` string. If nothing is specified the same model used during index creation is used. Available only if using Automated Embedding.
```
johnjackweir, line 1322: another field I don't see mentioned in docs

packages/mongodb-constants/src/stage-operators.ts
Anemy: Looking at it more, should they be commented to begin with?

packages/mongodb-constants/src/stage-operators.ts
```diff
@@ -1296,6 +1296,46 @@ const STAGE_OPERATORS = [
...
+ * path: The field to search. (Required)
+ * numCandidates: Number of nearest neighbors to use during the search. You can specify a number higher than the number of documents to return (\`limit\`) to increase accuracy. (Required)
+ * index: Name of the Atlas Search index to use. (Required)
+ * limit: Number (of type \`int\` only) of documents to return in the results. (Required)
+ * model: Specify a model, compatible with the one used during index creation, to perform embedding generation of the \`query\` string. If nothing is specified the same model used during index creation is used. Available only if using Automated Embedding. 
+i * filter: Any MongoDB Query Language (MQL) match expression that compares an indexed field with a boolean, number (not decimals), or string to use as a prefilter. (Optional)
+ * exact: Choose between false for ANN (Approximate Nearest Neighbor) and true for ENN (Exact Nearest Neighbor). Defaults to false. (Optional)
+ */
+`,
+  snippet: `{
```
johnjackweir, line 1327: there are a few other fields mentioned as optional in docs, should they also be added here?

johnjackweir (approved):
Nothing blocking but there are a few places that this diverges from docs, should this and or docs be updated?  https://www.mongodb.com/docs/vector-search/query/aggregation-stages/vector-search-stage/?deployment-type=atlas&embedding=byo&interface=driver&language=nodejs

packages/mongodb-constants/src/stage-operators.ts
```diff
@@ -1296,6 +1296,46 @@ const STAGE_OPERATORS = [
   },
 ] as const;
 
+const VECTOR_SEARCH_AUTO_EMBED_STAGE = {
+  name: '$vectorSearch',
```
Anemy, line 1300: We don't have any other duplicate stage operator definitions, I'm not sure how this will look in Compass, is there a reason we don't want to update the existing one?

packages/mongodb-constants/src/stage-operators.ts
mongodbnolanc: These additions are for auto embedding public preview and will be gated behind a 'preview' feature flag with an atlasCloudFeatureFlagName tie-in. My understanding is that have some docs changes in the works for this.

packages/mongodb-constants/src/stage-operators.ts
mongodbnolanc: The idea here is that we have a enableAutoEmbeddingPublicPreview feature flag that, if enabled, will replace the old $vectorSearch stage with the one defined here. Only one version will show at any given time, but we aren't ready to wholesale replace the old one.

packages/mongodb-constants/src/stage-operators.ts
mongodbnolanc: The idea here is that only one or the other gets used: `queryVector` if the index is bring-your-own-vector, `query` if the index is using auto embedding.

mongodbnolanc: Updated PR with new spacing

packages/mongodb-constants/src/stage-operators.ts
mongodbnolanc: Either query or queryVector are required, but only one of them, depending on which kind of vector search index is being queried against.

PR #727, opened 2026-05-05
feat(mongodb-ns): add explicit flag for internal dbs

gribnoysup:
See https://github.com/mongodb-js/compass/pull/8053

### DIFF
--- a/packages/mongodb-ns/src/index.spec.ts
+++ b/packages/mongodb-ns/src/index.spec.ts
@@ -96,6 +96,13 @@ describe('ns', function () {
     });
   });
 
+  it('should identify internal namespaces', function () {
+    assert.equal(ns('a').internal, false);
+    assert.equal(ns('a.b').isInternal(), false);
+    assert(ns('__mdb_internal').internal);
+    assert(ns('__mdb_internal_foo_bar').isInternal());
+  });
+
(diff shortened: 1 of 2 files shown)


packages/mongodb-ns/src/index.spec.ts
```diff
@@ -96,6 +96,13 @@ describe('ns', function () {
     });
   });
 
+  it('should identify internal namespaces', function () {
+    assert.equal(ns('a').internal, false);
+    assert.equal(ns('a.b').isInternal(), false);
```
nbbeeken, line 101:
```suggestion
    assert.equal(ns('a').internal, false);
    assert.equal(ns('a.b').isInternal(), false);
    assert.equal(ns('__mdb_internal.b').isInternal(), false);
    assert.equal(ns('__mdb_internal').isInternal(), false);
```

I recall when Jack originally worked on this I recall discussing that the trailing "_" was an important inclusion in the regex. Worth preventing a regression if some refactor comes by this code?

nbbeeken (approved):

packages/mongodb-ns/src/index.spec.ts
gribnoysup: Good idea, and thanks for the suggested edit!

PR #768, opened 2026-05-29
chore: update engines requirement in the repo

gribnoysup:
Preparing for trusted publishing, minimum age, allow scripts, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
     "email": "compass@mongodb.com"
   },
   "engines": {
-    "node": ">=14.17.5",
-    "npm": ">=9.0.0"
+    "node": ">=22.21.1",
+    "npm": ">=11.16.0"
   },
   "workspaces": [
     "packages/*",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -9,8 +9,8 @@
     "email": "compass@mongodb.com"
   },
   "engines": {
-    "node": ">=14.17.5",
```
gribnoysup, line 12: Tell me if you think there's a reason for us to keep it so low 🤔

nbbeeken (approved):

package.json
addaleax: For the top-level package.json? No, doesn't matter

For individual packages? If we think there's realistic external usage, bumping the minimum Node.js version is kind of a breaking change

gribnoysup: Yeah, no, this makes sense, but we only have a top-level engines in the monorepo, none of the packages specify their own. And we don't have versions that are lower in the test matrix, so if there are some requirements there, maybe we should add some for visibility

PR #769, opened 2026-06-01
feat(aggregation-stage-icons): add aggregation stage icons package

Anemy:
Having icons in the aggregation stage select options in Compass will help folks find their stages faster and add some visually descriptive information for the stages. I'm opening the pr here instead of Compass as I figure as the icons could be used in another place down the line, like VSCode. If folks feel it's better in Compass let me know! Happy to move it over.

I haven't made an icon react package library before, let me know if there are any suggestions around exporting/typing. Here's [LG's icon package](https://github.com/mongodb-js/leafygreen-ui/tree/main/packages/icon) as a reference.

This list of icons isn't designed to be an exhaustive list of all of the stages. The consumer will have to have fallbacks for any unsupported stages. 

Misba designed these! With some AI assistance along the way.

Here's how it will look in Compass:
Screenshot 2026-05-29 at 04 53 47

All of the icons  * **some of them have been tweaked since this screenshot was taken**
Screenshot 2026-06-01 at 17 53 28


nbbeeken (approved):

addaleax (approved):

gribnoysup (approved):
Looks good, but wouldn't it make more sense to just have them close to constants so that we know when to add a new one?

Anemy:
@gribnoysup I was thinking about adding it there and decided against it just to keep react out of mongodb-constants. I don't feel too strongly so I'm happy moving it there if you'd like. Eventually I think it makes the most sense to be in the design systems. I've asked about that in their slack channel https://mongodb.slack.com/archives/CFFA5BS79/p1780416541946559

gribnoysup:
FWIW I don't think we need to package React with it like that, svg paths are already very portable, but no strong opinions on that

PR #772, opened 2026-06-02
fix(devtools-proxy-support): reconnect SSH tunnel after unrecoverable client error COMPASS-8355

ivandevp:
## Description

Fixes the SSH tunnel failing to recover after a laptop resumes from hibernate (COMPASS-8355).

When the OS hibernates, the underlying TCP connection is killed mid-stream. The ssh2 packet parser receives truncated data and puts the `Client` instance into a permanently broken "unusable" state. Two problems followed:

1. There was no `sshClient.on('error', ...)` handler in `SSHAgent`, so the `'error'` event emitted by the broken parser went completely unhandled → uncaught exception crashing the process.
2. The existing retry logic in `_connect()` only matched `"Not connected"` and `"Channel open failure"`. Even if `"Instance unusable after fatal error"` had been added to that list, the retry would have called `initialize()` on the **same broken instance** → same error again. The ssh2 `Client` must be recreated before reconnecting.

### Changes

**`createSshClient()` (new private method)** — extracts ssh2 `Client` construction and listener setup into a reusable method. Adds a persistent `client.on('error', ...)` handler that sets `connected = false` and prevents unhandled error events from crashing the process.

**`initialize(reinitializeClient = false)`** — the new flag triggers `sshClient.end()` + `createSshClient()` before reconnecting, discarding the broken instance and binding `forwardOut` to the fresh one.

**`_connect()`** — expands retryable error patterns to cover `"Instance unusable after fatal error"`, `"read ECONNRESET"`, and `"Socket closed"`. When `requiresNewClient` is true, passes the flag through to `initialize()`.

### New tests

- `'does not crash on unexpected sshClient error events'` — verifies the new `'error'` handler absorbs the event without throwing.
- `'reconnects with a fresh SSH client after "Instance unusable after fatal error"'` — simulates a broken ssh2 client by patching `connect()` to emit that error, then verifies a successful reconnect with a fresh client (auth handler called twice).

## Open Questions

This is step 1 of a broader cleanup (see COMPASS-8355). Step 2 will update `devtools-connect` to expose the tunnel reference or surface tunnel errors on `DevtoolsConnectionState`, so that Compass no longer needs to create and manage the tunnel itself.

## Checklist

### DIFF
--- a/packages/devtools-proxy-support/src/ssh.spec.ts
+++ b/packages/devtools-proxy-support/src/ssh.spec.ts
@@ -2,7 +2,6 @@ import { HTTPServerProxyTestSetup } from '../test/helpers';
 import { SSHAgent } from './ssh';
 import { createFetch } from './fetch';
 import { expect } from 'chai';
-import { once } from 'events';
 import sinon from 'sinon';
 
 describe('SSHAgent', function () {
--- a/packages/devtools-proxy-support/src/ssh.spec.ts
+++ b/packages/devtools-proxy-support/src/ssh.spec.ts
@@ -125,4 +125,46 @@ describe('SSHAgent', function () {
     expect(setup.authHandler).to.have.been.calledTwice;
     expect(setup.canTunnel).to.have.been.calledTwice;
   });
+
+  it('does not crash on unexpected sshClient error events', async function () {
+    agent = new SSHAgent({
+      proxy: `ssh://foo:bar@127.0.0.1:${setup.sshProxyPort}/`,
+    });
+    await agent.initialize();
+    expect(() => {
+      (agent as any).sshClient.emit(
+        'error',
+        new Error('some unexpected ssh error'),
+      );
+    }).not.to.throw();
+    expect((agent as any).connected).to.be.false;
+  });
+
+  it('reconnects with a fresh SSH client after "Instance unusable after fatal error"', async function () {
+    setup.authHandler = sinon.stub().returns(true);
+    agent = new SSHAgent({
+      proxy: `ssh://foo:bar@127.0.0.1:${setup.sshProxyPort}/`,
+    });
+    await agent.initialize();
+    const fetch = createFetch(agent);
+    await fetch('http://example.com/hello');
+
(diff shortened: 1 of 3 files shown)


packages/devtools-proxy-support/src/ssh.spec.ts
```diff
@@ -125,4 +125,46 @@ describe('SSHAgent', function () {
...
+    agent = new SSHAgent({
+      proxy: `ssh://foo:bar@127.0.0.1:${setup.sshProxyPort}/`,
+    });
+    await agent.initialize();
+    expect(() => {
+      (agent as any).sshClient.emit(
+        'error',
+        new Error('some unexpected ssh error'),
+      );
+    }).not.to.throw();
```
addaleax, line 140: We can keep this test, but I think a much more valuable test would be one in which the actual connection is interrupted – and luckily, the setup here _does_ allow us to test this directly, because `HTTPServerProxyTestSetup` holds a reference to an SSH server, and that SSH server forwards `getConnections` to the underlying `net.Server`, and we can call `.destroy()` on those connections.

packages/devtools-proxy-support/src/ssh.ts
```diff
@@ -75,6 +87,16 @@ export class SSHAgent extends AgentBase implements AgentWithInitialize {
       throw new Error('Disconnected.');
     }
 
+    if (reinitializeClient) {
+      // The previous ssh2 Client instance is in an unrecoverable state (e.g.
+      // "Instance unusable after fatal error" after the TCP connection was killed
+      // mid-stream during hibernate). Create a fresh client before reconnecting.
+      delete this.connectingPromise;
```
addaleax:
```suggestion
      this.connectingPromise = undefined;
```

(not that important, but `delete` is generally more expensive in JS when not targeting dictionary objects)

packages/devtools-proxy-support/src/ssh.spec.ts
ivandevp: test added :)

addaleax (approved):

PR #778, opened 2026-06-05
feat(tracking-plan): generate tracking plan from source MONGOSH-3405

alenakhineika:
## Description

Extracts [generate-tracking-plan.ts](https://github.com/mongodb-js/compass/blob/main/scripts/generate-tracking-plan.ts) from Compass and applies some refactoring so it can be used by different products (Compass as well as mongosh). The products will require a small amount of changes to work with the updated script.

## Checklist


packages/tracking-plan/src/index.ts
addaleax, line 1: This package isn't being published, it's meant for scripts that are part of the monorepo management itself – you'll probably just want to create a new package for this

addaleax (approved):

PR #784, opened 2026-06-17
fix(shell-bson-parser): do not allow arbitrary function calls on literals COMPASS-10678

gribnoysup:
We are already pretty strict about allowed function calls during parse, this patch makes the logic a bit stricter. I can't think of a case where we'd want to allow these cases for any of our purposes, but tell me if we should allowlist some methods here

See COMPASS-10678 for some extra context


addaleax (approved):
Seems like strictly an improvement, ~~but the source change here isn't tested, right?~~ nvm, just missed that it's rolled into the same test case

gribnoysup:
Whoops, my bad, should've split it in two separate ones

PR #788, opened 2026-06-19
chore(tracking-plan): add LICENSE; don't publish dev-only files; mark package as public

gribnoysup:
New package was not set up fully for publishing

### DIFF
--- a/packages/tracking-plan/package.json
+++ b/packages/tracking-plan/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mongodb-js/tracking-plan",
-  "version": "0.2.0",
+  "version": "0.3.0",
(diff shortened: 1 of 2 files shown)


packages/tracking-plan/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mongodb-js/tracking-plan",
-  "version": "0.2.0",
+  "version": "0.3.0",
```
gribnoysup, line 3: I published 0.2.0 first, but because `files` was missing in package.json, it published all the dev-only files, had to unpublish it, but npm wouldn't allow me to publish the same version again (probably caching, but I didn't want to wait), so I bumped it

PR #805, opened 2026-06-27
fix(tracking-plan): identifyTraits must have payload

alenakhineika:
## Description

The `IdentifyTraits` interface is a Segment-specific concept. As part of a mongosh telemetry work, we are moving away from Segment, identify calls are now expressed as regular TelemetryEvent members with a payload field, consistent with all other events. Compass can adopt the same pattern by wrapping its existing IdentifyTraits properties inside a payload field, keeping the shared module working for both tools without any special-casing.

## Checklist


addaleax (approved):
This has externally visible impact on this package so should be `fix():` rather than `refactor():`

PR #806, opened 2026-06-29
feat: expose array index in $map, $filter, $reduce for MongoDB 8.3 MONGOSH-3244

ivandevp:
## Summary

- Adds `$$IDX` system variable to `mongodb-constants`, version-gated to MongoDB 8.3+ (used by Compass filter autocomplete)
- Adds `arrayIndexAs` optional field to `$filter` and `$map` operator types in `mql-typescript/out/schema.d.ts`
- Adds `as`, `valueAs`, and `arrayIndexAs` optional fields to `$reduce` operator types (all new in MongoDB 8.3)
- These schema changes feed into `mongodb-ts-autocomplete`'s `autocomplete-types.ts` (auto-generated during build), which powers the TS-based autocompleter in mongosh and Compass

## Motivation

MongoDB 8.3 extends `$map`, `$filter`, and `$reduce` to expose the index of the current array element via a new `arrayIndexAs` field (defaulting to `$$IDX`). `$reduce` also gains `as` and `valueAs` to rename the `$$this` and `$$value` variables respectively. Tracked in [MONGOSH-3244](https://jira.mongodb.org/browse/MONGOSH-3244).

## Test plan

- [x] `mongodb-constants` tests: 123 passing (added 2 tests for `$$IDX` version gating — present on 8.3+, absent on 8.2-)
- [x] `mongodb-ts-autocomplete` tests: 12 passing (added 1 test that `arrayIndexAs` is suggested inside `$filter`, `$map`, `$reduce` operator bodies)
- [ ] After merge: bump `@mongodb-js/mongodb-constants` and `@mongodb-js/mongodb-ts-autocomplete` in the mongosh repo to pick up the published versions


packages/mql-typescript/out/schema.d.ts
```diff
@@ -1560,6 +1560,12 @@ export namespace Aggregation.Expression {
        */
       as?: string;
 
+      /**
+       * A name for the variable that represents the index of the array element. Available in MongoDB 8.3+.
+       * If not specified, the index is available as $$IDX.
+       */
+      arrayIndexAs?: string;
```
addaleax, line 1567: This is a generated file. We're generating them from YAML files; originally they were part of the PHP driver repository, now they are maintained separately in the mql-specifications repo.

I think this PR might have a dependency on https://github.com/mongodb-js/devtools-shared/pull/622, unless you want to update the PHP driver branch manually.

ivandevp: oh, I see - that makes sense. What are we missing to get #622 merged?

addaleax: I'm not fully remembering, but locally it seems some tests are failing. That probably requires some coordination with the upstream repository – I can try to take a deeper look tomorrow, but there's a good chance you'll need to look into these as well

addaleax: I think the order of operations is:

1. Get https://github.com/mongodb/mql-specifications/pull/53 merged
1. Then, switch to the mql-specifications repo main branch in https://github.com/mongodb-js/devtools-shared/pull/622 instead of the branch for the above PR
1. Get https://github.com/mongodb-js/devtools-shared/pull/622 merged
1. Add content from this PR to mql-specifications, open a PR, merge it
1. Adopt the latest version of mql-specifications in this repository

PR #822, opened 2026-07-15
feat: DSC support for mongodb runner

mpobrien:
Adds the ability to launch DSC clusters backed by SLS.

- New `disaggregatedStorage` cluster option (all topologies): starts a docker compose project for the storage backend, waits for readiness, runs per-shard setup, and injects `disaggregatedStorageConfig`/`disaggregatedStorageEnabled` into every mongod. Compose project is shared across shards and torn down with the cluster (including on failed startup).
- SLS helpers: `createSLSDisaggregatedStorageOptions()` handles the full flow for the SLS multi-cell compose file from a server checkout — env/port allocation (services parsed from the compose file itself), readiness polling, per-shard `StartLog`, encryption key file, and auto-generated `disaggregatedStorageConfig` (incl. `replSetConfig` from pre-allocated member ports).
- DSC-specific behavior: replset defaults to 2 nodes, `replSetInitiate` is skipped (config comes from startup parameters), and `local`-db metadata tracking is disabled (not writable on DSC).
- New CLI flags: `--slsCompose`/`--slsImageTag` (full SLS flow), `--disaggregatedStorageCompose`/`--disaggregatedStorageConfig` (custom backends), and `--downloadUrl` for using a direct mongod tarball URL
  (also added to mongodb-downloader, cached by URL).
- Misc: docker compose output + download progress in debug logs, `mongodb-runner start` prints allocated SLS service ports.

Sample invocation:
```
node bin/runner.js start -t replset \
  --slsCompose=/home/ubuntu/mongo/buildscripts/modules/atlas/sls-multicell-docker-compose.yml \
  --slsImageTag=dfb12c1092695ae79f6ff3c5b6c31ff7417e4843 \
  --binDir=/mnt/mac/Users/mikeo/ubuntu-mongod/dist-test/bin/ \
  --debug --logDir=/tmp/mongodb-runner-logs
```

## Checklist


packages/mongodb-runner/src/oidc.js
```diff
@@ -0,0 +1,113 @@
+import { spawn } from 'child_process';
+import { once } from 'events';
+import { parseCLIArgs, OIDCMockProvider } from '@mongodb-js/oidc-mock-provider';
+import { debug } from './util';
+if (process.env.RUN_OIDC_MOCK_PROVIDER !== undefined) {
+  (async function main() {
+    const uuid = crypto.randomUUID();
+    debug('starting OIDC mock provider with UUID', uuid);
```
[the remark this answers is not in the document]
mpobrien: Fixed in ecef1cf — the compiled .js files in src/ were committed accidentally and have been removed; the .ts source now imports randomUUID explicitly from 'crypto'.

packages/mongodb-runner/src/oidc.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+  async close() {
+    try {
+      if (this.pid) process.kill(this.pid, 0);
+    } catch (e) {
+      if (typeof e === 'object' && e && 'code' in e && e.code === 'ESRCH')
+        return; // process already exited
+    }
+    if (!this.issuer || !this.uuid) return;
+    await fetch(new URL(this.issuer, `/shutdown/${this.uuid}`));
```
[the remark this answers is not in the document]
mpobrien: Fixed in ecef1cf — arguments swapped to `new URL(`/shutdown/${uuid}`, this.issuer)` in oidc.ts (the stale compiled .js files were removed).

packages/mongodb-runner/src/util.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+export async function safePromiseAll(promises) {
+  const results = await Promise.allSettled(promises);
+  const rejected = results.filter((r) => r.status === 'rejected');
+  if (rejected.length) {
+    if (rejected.length === 1) throw rejected[0].reason;
+    throw new AggregateError(
+      [rejected.map((r) => r.reason)],
+      `${rejected.length} errors: ${rejected.map((r) => r.reason).join(', ')}`,
+    );
```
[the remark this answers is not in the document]
mpobrien: Fixed in ecef1cf — the errors array is no longer nested; `AggregateError.errors` now contains the individual errors.

packages/mongodb-runner/src/mongocluster.ts
```diff
@@ -420,6 +607,41 @@ export class MongoCluster extends EventEmitter<MongoClusterEvents> {
       );
     } else if (options.topology === 'replset') {
       const { rsMembers, replSetName } = processRSMembers(options);
+      if (disaggregatedStorage) {
+        // Pre-allocate member ports so the storage configuration (which
+        // embeds the replica set config, including member host:ports) can be
+        // computed before the servers start.
+        const members: ShardMemberDescriptor[] = [];
```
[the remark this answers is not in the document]
mpobrien: Fixed in ecef1cf — `ShardMemberDescriptor` now carries `arbiterOnly`, populated from the replica set member options, and the generated DSC `replSetConfig` includes it for each member.

packages/mongodb-runner/src/mongocluster.js
```diff
@@ -0,0 +1,542 @@
...
+      defaultConnectionOptions: jsonClone(this.defaultConnectionOptions ?? {}),
+      users: jsonClone(this.users),
+    };
+  }
+  isClosed() {
+    // Return true if and only if there are no running sub-clusters/servers
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    for (const _ of this.children()) return true;
+    return true;
+  }
```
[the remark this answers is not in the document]
mpobrien: Fixed in ecef1cf — `isClosed()` now returns false while child servers/shards exist (the stale compiled .js copy was removed).

packages/mongodb-downloader/src/index.ts
```diff
@@ -31,6 +32,11 @@ export type MongoDBDownloaderOptions = {
   useLockfile: boolean;
   /** The options to pass to the download URL lookup. */
   downloadOptions?: DownloadOptions;
+  /**
+   * A direct URL to a MongoDB tarball to download. If set, `version` and
+   * `downloadOptions` are ignored and no download URL lookup is performed.
+   */
+  downloadUrl?: string;
```
gribnoysup, line 54: I think we should think about some different approach for adding this functionality in the downloader: the issue I see here is that while downloadOptions will be ignored, we will still consider them somewhat when returning the DownloadResult metadata, which from the consumer perspective might put you in a weird state where what you download doesn't match the returned options at all

packages/mongodb-downloader/src/index.ts
```diff
@@ -99,11 +114,13 @@ export class MongoDBDownloader {
           }
 
           await fs.mkdir(downloadTarget, { recursive: true });
-          const artifactInfo = await this.lookupDownloadUrl({
-            targetVersion: version,
-            enterprise: isEnterprise,
-            options: downloadOptions,
-          });
+          const artifactInfo = downloadUrl
+            ? ({ url: downloadUrl } as DownloadArtifactInfo)
```
gribnoysup: This is a dangerous assertion that makes the public interface here return a value that doesn't match the provided type at all and can cause consumer code to crash completely

packages/mongodb-runner/src/cli.ts
```diff
@@ -114,7 +138,38 @@ import type { MongoClientOptions } from 'mongodb';
   }
 
   async function start() {
-    const { cluster, id } = await utilities.start(argv, args);
+    if (argv.slsCompose && !argv.slsImageTag) {
+      throw new Error('--slsCompose requires --slsImageTag');
+    }
```
gribnoysup: With the `yargs` library you can specify these dependencies in the options definition

packages/mongodb-runner/src/cli.ts
```diff
@@ -114,7 +138,38 @@ import type { MongoClientOptions } from 'mongodb';
...
+          imageTag: argv.slsImageTag!,
+        })
+      : argv.disaggregatedStorageCompose
+        ? {
+            composeFile: argv.disaggregatedStorageCompose,
+            config: (() => {
+              try {
+                return JSON.parse(argv.disaggregatedStorageConfig ?? '');
+              } catch {
+                return argv.disaggregatedStorageConfig ?? '';
```
gribnoysup: Should this continue if we couldn't parse the arg?

packages/mongodb-runner/src/mongocluster.ts
```diff
@@ -369,9 +484,16 @@ export class MongoCluster extends EventEmitter<MongoClusterEvents> {
...
-  }: MongoClusterOptions): Promise<MongoCluster> {
+  static async start(
+    { ...options }: MongoClusterOptions,
+    // Internal parameter used when starting the shards of a sharded cluster:
+    // the parent cluster owns the docker compose project, and the shards
+    // inherit the DSC context from it.
+    _internal?: {
+      disaggregatedStorage: DisaggregatedStorageOptions;
+      shardContext: { index: number; isConfigServer: boolean };
+    },
```
gribnoysup: I find it a bit clunky that we're expressing this logic via an "internal" parameter on a public interface. If the idea here is that we want to know that the `MongoCluster` is a child of another `MongoCluster` that might be a disagg one, maybe we can model it that way and just pass a parent cluster as a second param here (and expose whatever we need from the parent so that the child cluster can read it)

packages/mongodb-runner/src/mongocluster.ts
```diff
@@ -411,7 +534,79 @@ export class MongoCluster extends EventEmitter<MongoClusterEvents> {
...
+          // Only the compose-owning (top-level) cluster waits for readiness.
+          await disaggregatedStorage.waitForReady?.();
+        }
+        if (options.secondaries === undefined) options.secondaries = 1;
+      }
+      return await this._startServers(
+        cluster,
+        options,
+        disaggregatedStorage,
+        _internal?.shardContext ?? { index: 0, isConfigServer: false },
```
gribnoysup: This fallback value is kinda weird: it looks like it's generic enough to apply to all shards we create, but it's gated by disagg check, so if someone reads it while creating a non-disagg cluster it will actually report state that is contradictory (a config server we're trying to start will read `isConfig: false` for example). Can we either rename this and make it optional to make sure it's clear it's disagg-only, or just make it generic enough so that the value is correct for all cases?

packages/mongodb-runner/src/cli.ts
```diff
@@ -78,6 +83,25 @@ import type { MongoClientOptions } from 'mongodb';
...
+    })
+    .option('disaggregatedStorageCompose', {
+      type: 'string',
+      describe: 'Path to docker-compose.yml for the DSC backend',
+    })
+    .option('disaggregatedStorageConfig', {
+      type: 'string',
+      describe:
+        'JSON value for the disaggregatedStorageConfig setParameter on each mongod',
+    })
```
gribnoysup: If you should either provide sls or disaggregated args (based on the logic I see below), maybe worth validating this case aborting if all these are provided?

gribnoysup (commented):
The high level direction this is going looks good to me. I didn't do a thorough review yet as this is a draft, but a couple of things already caught my eye, so I left a couple of comments here and there. Happy to discuss those more

mpobrien:
Thanks for the in-depth review @gribnoysup ! Will work through these, I might have a few follow-up questions.

packages/mongodb-runner/src/cli.ts
mpobrien: nice, added the `implies` field in the `yargs` options to do this instead.

autarch:
Is there anything I can help with here? I need this to add DSC tests to the DB Tools.

packages/mongodb-downloader/src/index.ts
mpobrien: removed the cast; added a DownloadUrlResult type (Partial<DownloadArtifactInfo> & { url, downloadedBinDir }) and overloads on downloadMongoDbWithVersionInfo(), so URL-based downloads advertise only what's guaranteed

packages/mongodb-runner/src/mongocluster.ts
mpobrien: Moved the implementation for this into private `static _start(options, internal)`.  `public start(options)` takes one argument again, and shard recursion goes through the private method

packages/mongodb-runner/src/mongocluster.ts
mpobrien: Changed this so that shard context is now always passed down for shards (not just in DSC mode), so the { index: 0, isConfigServer: false } default only applies to top-level clusters, where it's accurate.

packages/mongodb-runner/src/cli.ts
mpobrien: Ended up stripping these out since there's no real reason to expose `disaggregatedStorageConfig`, the correct value for each mongod gets generated internally and there's likely no real use case to need to override it.
`disaggregatedStorageCompose` is redundant with `slsCompose`, so I just dropped it. With the remaining set of args, its safe/valid for them to all coexist, so no additional validation should be needed.

mpobrien:
@gribnoysup sorry for the delay, I addressed the comments.

mpobrien:
@autarch until this gets merged you can check out the branch and use that to get started on using it for tools testing?

packages/mongodb-runner/src/mongocluster.ts
```diff
@@ -369,9 +484,24 @@ export class MongoCluster extends EventEmitter<MongoClusterEvents> {
...
+  }
+
+  // Also used when starting the shards of a sharded cluster: shardContext
+  // identifies the shard being started, and, if the parent cluster owns a
+  // DSC docker compose project, the shards inherit the DSC context from it.
+  // The shardContext default below is only applied to top-level (non-shard)
+  // clusters, for which it is accurate.
+  private static async _start(
+    { ...options }: MongoClusterOptions,
+    internal:
```
gribnoysup: Can we give this a better name? Internal doesn't tell much when reading through the code. I guess this is like `parentClusterContext` or something?

packages/mongodb-downloader/src/index.ts
```diff
@@ -39,7 +58,8 @@ export class MongoDBDownloader {
     version = '*',
     directory,
     useLockfile,
-  }: MongoDBDownloaderOptions): Promise<DownloadResult> {
+    downloadUrl,
+  }: MongoDBDownloaderOptions): Promise<DownloadResult | DownloadUrlResult> {
```
gribnoysup, line 73: We should add the same function overloads as you added below, otherwise it changes the interface here in a way that would require every user to always manually assert whether or not the result contains expected values

packages/mongodb-downloader/src/index.ts
```diff
@@ -139,11 +170,35 @@ export class MongoDBDownloader {
     const response = await fetch(url, {
       highWaterMark: MongoDBDownloader.HWM,
     } as Parameters<typeof fetch>[1]);
+    if (!response.ok) {
+      throw new Error(
+        `Failed to download ${url}: ${response.status} ${response.statusText}`,
+      );
+    }
+    const totalBytes = +(response.headers.get('content-length') ?? '');
+    const totalMB = totalBytes ? (totalBytes / 1048576).toFixed(1) : null;
```
gribnoysup: Nit: can we move the number to a var so that it's very clear immediately what it is and we always consistently use the same number below?

gribnoysup (changes requested):
A couple of small things, but otherwise looks good to me. I couldn't test the new cluster type myself locally, but at the very least the package tests are passing and using the new version in devtools repos also still runs all the tests correctly, so that's a good sign

autarch:
> @autarch until this gets merged you can check out the branch and use that to get started on using it for tools testing?

Yes, that's what I'm doing.

packages/mongodb-downloader/src/index.ts
mpobrien: sure, done

packages/mongodb-runner/src/mongocluster.ts
mpobrien: 👍

mpobrien:
Addressed the comments.
There's some errors in the checks that look unrelated to this PR? 
```
Run scopes=$(node .github/scripts/compute-test-group.mjs "1" "${TOTAL_GROUPS}")
  scopes=$(node .github/scripts/compute-test-group.mjs "1" "${TOTAL_GROUPS}")
  echo "Testing group 1/${TOTAL_GROUPS} with: ${scopes}"
  npm run test-ci -- --stream ${scopes}
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    TOTAL_GROUPS: 4
    SINCE_REF: remotes/origin/main
    pythonLocation: /opt/hostedtoolcache/Python/3.11.15/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.15/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.15/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.15/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.15/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.15/x64/lib
node:internal/modules/cjs/loader:1386
  throw err;
  ^

Error: Cannot find module '/home/runner/work/devtools-shared/devtools-shared/.github/scripts/compute-test-group.mjs'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1025:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1030:22)
    at Function._load (node:internal/modules/cjs/loader:1192:37)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:237:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

nbbeeken:
@mpobrien It is sort of a quirk of github actions, the workflow is defined on the main branch but the code is checked out to this branch so the script file that the workflow calls away to is missing. This branch just needs a rebase or merge-in of main

mpobrien:
@nbbeeken ack, thanks - i rebased on main, let's see if that takes care of it.

gribnoysup (approved):

PR #829, opened 2026-07-23
chore: bump mongodb and bson dependencies

nbbeeken:
## Description

Bump `mongodb` and `bson` across packages and fix where script to do so

## Checklist


lerouxb (approved):

.github/workflows/check-test.yaml
```diff
@@ -25,10 +25,16 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
+        # Tests are sharded across hosts by package (see compute-test-shard.mjs).
+        # When changing the shard count, update every list entry below.
+        shard: [1, 2, 3, 4]
```
Anemy: nit, not a blocker. Should we use the wording `group` instead of `shard` like we do for Compass e2e tests? https://github.com/mongodb-js/compass/blob/80ef906493ef0891197fdb4c9af324c1d98d2e43/.evergreen/buildvariants-and-tasks.in.yml#L162 
I'm worried `shard` could cause confusion. Maybe there's a better word?

nbbeeken: Terminology alignment totally worth doing, group is good, fixed!

Anemy (approved):

.github/workflows/compute-test-group.mjs
Anemy, line 1: Is there a reason to not use typescript here?

nbbeeken: Node.js 22 is currently the test node here so it would need a flag to run the ts, I could do the flag, or I could upgrade the node, I chose neither 🤣

PR #851, opened 2026-08-03
fix(devtools-proxy-support): support SSH password auth via `keyboard-interactive` COMPASS-10929

### LINKED ISSUE #839: SSH proxy (tunnel) password authentication fails with `keyboard-interactive`-only servers
#### Current Behavior

MongoDB Compass fails to connect through its built-in SSH tunnel when the SSH server accepts a password through `keyboard-interactive`/PAM but does not advertise the separate SSH `password` authentication method.

Compass reports:

```text
All configured authentication methods failed
```

The failure occurs during SSH authentication, before Compass attempts to connect to MongoDB.
The same credentials work in SSH clients that use the configured password to answer a keyboard-interactive password prompt.

The issue can be reproduced as follows:

1. Configure an OpenSSH server to offer `keyboard-interactive` authentication while
   `PasswordAuthentication` is disabled.
2. Verify that the server advertises:

   ```text
   publickey,keyboard-interactive
   ```

3. Run MongoDB on the SSH host with its listener bound to `127.0.0.1`.
4. Configure a Compass connection using the built-in SSH tunnel and the existing SSH `Password` mode.
5. Enter valid SSH credentials and attempt to connect.

#### Code/Gist

`SSHAgent` passes the decoded password to `ssh2.Client.connect()`:

[`packages/devtools-proxy-support/src/ssh.ts`](https://github.com/mongodb-js/devtools-shared/blob/6a47418fa242eb063475f7c24729a40de0cb4c0f/packages/devtools-proxy-support/src/ssh.ts#L111-L119)

However, each `ssh2.Client` is created without a `keyboard-interactive` event handler:

[`SSHAgent.createSshClient()`](https://github.com/mongodb-js/devtools-shared/blob/6a47418fa242eb063475f7c24729a40de0cb4c0f/packages/devtools-proxy-support/src/ssh.ts#L62-L70)

The [`ssh2` 1.17.0 documentation](https://github.com/mscdex/ssh2/blob/v1.17.0/README.md) requires `tryKeyboard: true` together with a `keyboard-interactive` event handler.
Supplying `password` alone configures only the distinct SSH `password` authentication method.

#### Expected Behavior/Code

Compass's existing SSH `Password` mode should support both the regular SSH `password` method and a conventional keyboard-interactive/PAM password challenge containing one hidden prompt.

The stored password should be sent only to the first non-empty round when that round contains exactly one prompt with `echo === false`.
It should not be sent to visible prompts, multiple prompts, or later rounds that may request an OTP or another authentication factor.

No new Compass setting or public API is required.

#### Environment

- Node.js 22.23.1 / npm 11.17.0
- OS: Windows Compass client; OpenSSH server advertising `publickey,keyboard-interactive`
- MongoDB Compass: 1.49.12
- `@mongodb-js/devtools-proxy-support`: 0.7.17 in the packaged application
- `ssh2`: 1.17.0
- MongoDB listener on the SSH host: `127.0.0.1:27017`

#### Possible Solution

When the SSH URL contains a password:

1. Set `tryKeyboard: true` in the `ssh2` connection configuration.
2. Register a `keyboard-interactive` handler for each new `ssh2.Client`.
3. Answer zero-prompt informational rounds with `[]` without consuming the password attempt.
4. Send the configured password only to the first non-empty round containing one hidden prompt.
5. Return empty responses for multiple prompts, visible prompts, and subsequent non-empty rounds.
6. Keep the challenge state scoped to the client so a newly created client receives fresh state.

Regular SSH `password` authentication should continue to receive the same decoded password.
General interactive conversations and MFA remain outside this compatibility fix.

#### Additional Context/Screenshots

A runtime-only patch (extracted `app.asar`) implementing the proposed behavior was tested against Compass 1.49.12:

- the unmodified application failed while the server advertised `publickey,keyboard-interactive`;
- the patched application established the SSH tunnel and connected to MongoDB;
- an incorrect SSH password was rejected without hanging;
- a fully restarted Compass process connected successfully again;
- regular SSH `password` authentication continued to work after it was re-enabled on the server.

Two community reports describe the same Compass error while using an SSH proxy:

- [Getting error “All configured authentication methods failed” while connecting Mongo Atlas from Compass 1.44 using SSH proxy](https://www.mongodb.com/community/forums/t/getting-error-all-configured-authentication-methods-failed-while-connecting-mongo-atlas-from-compass-1-44-using-ssh-proxy-with/295590)
- [MongoDB Compass not able to connect using proxy SSH with password](https://www.mongodb.com/community/forums/t/mongodb-compass-not-able-to-connect-using-proxy-ssh-with-password/295637)

Those reports do not include the server's advertised SSH authentication methods, so they may share the symptom without necessarily having the same root cause.

TemaSM:
Fixes mongodb-js/devtools-shared#839

## Description

MongoDB Compass's SSH `Password` mode passes the configured password to `ssh2`, but does not opt in to `keyboard-interactive` authentication.  
This makes SSH tunnels fail when an OpenSSH/PAM server accepts that password only through `keyboard-interactive`.

The SSH agent now enables `tryKeyboard` when a password is configured and uses it for one hidden, single-prompt challenge.  
Empty rounds do not consume the attempt, while visible prompts, multiple prompts, and later rounds receive empty responses so the stored password is not reused for OTP or MFA prompts.

A per-connection authentication selector preserves the existing `none → password → publickey → keyboard-interactive` order.  
Once the server reports `partialSuccess`, the keyboard-interactive password fallback is skipped for that connection, while other configured methods such as `publickey` remain available.  
A reconnect starts with fresh authentication state.

This does not change the public API, connection options, or dependencies.

### Root cause

`SSHAgent` passed `password` to `ssh2.Client.connect()`, but did not enable `tryKeyboard` or register the `keyboard-interactive` event handler required by `ssh2`.  
Servers advertising `publickey,keyboard-interactive` therefore exhausted the configured authentication methods before Compass attempted to connect to MongoDB, showing user popup with text: `All configured authentication methods failed`.

The initial password fallback also needed to account for `partialSuccess`.  
The default `ssh2` method selector does not use that flag, so a server could accept `password` or `publickey` as one factor and then receive the configured password in its first hidden OTP prompt.

### Changes

- Enable `tryKeyboard` only when the SSH URL contains a password.
- Answer only the first non-empty keyboard-interactive round containing exactly one hidden prompt.
- Return empty responses for visible prompts, multiple prompts, and later non-empty rounds.
- Keep the prompt state scoped to one SSH connection and reset it when a reusable client socket closes.
- Use an internal per-connection authentication selector that keeps the supported order `none → password → publickey → keyboard-interactive`.
- Track `partialSuccess` cumulatively and skip the keyboard-interactive password fallback after any method succeeds partially.
- Keep other configured methods, including `publickey`, available after a partial success.
- Accept `methodsLeft` without using it to reorder methods in this change.
- Extend the local `ssh2.Server` fixture and add unit and integration coverage for password, public-key, partial-success, prompt-shape, rejection, and reconnection paths.

### Security boundary

This is a password compatibility fallback, not a general interactive or MFA implementation.  
Before any authentication method succeeds, the configured password is eligible only for the first non-empty round containing one hidden prompt.

After `ssh2` reports `partialSuccess`, keyboard-interactive is skipped for the rest of that SSH connection.  
A later failed method cannot clear that state, while a new connection receives a fresh selector.

### Testing & validation

- Added unit tests for authentication method selection and `partialSuccess` handling.
- Extended the local `ssh2.Server` fixture to cover password, public-key, keyboard-interactive, and multi-step authentication.
- Covered empty, visible, multiple, repeated, rejected, and reconnecting keyboard-interactive scenarios.
- Targeted tests: 38 passing.
- Full package suite: 85 passing, 3 pending.
- TypeScript compilation and package checks pass.

Tested with Node.js 22.23.1 and npm 11.17.0:

- `npm run test --workspace=@mongodb-js/devtools-proxy-support -- src/ssh-auth.spec.ts src/ssh.spec.ts`: 38 passing
- `npm run compile --workspace=@mongodb-js/devtools-proxy-support`: passing
- `npm run check --workspace=@mongodb-js/devtools-proxy-support`: passing
- `npm run test-ci --workspace=@mongodb-js/devtools-proxy-support`: 85 passing, 3 pending, 0 failing

The basic `keyboard-interactive` password path was also tested manually with a locally patched Compass 1.49.12 build against a server advertising `publickey,keyboard-interactive`.
The `partialSuccess` boundary is covered by the automated SSH server fixture.

## Open Questions

None.

## Checklist

~- [ ] Documentation is changed or added~

### DIFF
--- a/packages/devtools-proxy-support/src/ssh-auth.spec.ts
+++ b/packages/devtools-proxy-support/src/ssh-auth.spec.ts
@@ -1,0 +1,180 @@
+import { expect } from 'chai';
+import { createSshAuthMethodSelector } from './ssh-auth';
+
+describe('createSshAuthMethodSelector', function () {
+  it('only attempts none without configured credentials', function () {
+    const selectAuthMethod = createSshAuthMethodSelector({
+      hasPassword: false,
+      hasPrivateKey: false,
+    });
+
+    expect(selectAuthMethod(null, null)).to.equal('none');
+    expect(selectAuthMethod([], false)).to.equal(false);
+  });
+
+  it('preserves the password authentication order', function () {
+    const selectAuthMethod = createSshAuthMethodSelector({
+      hasPassword: true,
+      hasPrivateKey: false,
+    });
+
+    expect(selectAuthMethod(null, null)).to.equal('none');
+    expect(
+      selectAuthMethod(['password', 'keyboard-interactive'], false),
+    ).to.equal('password');
+    expect(selectAuthMethod(['keyboard-interactive'], false)).to.equal(
+      'keyboard-interactive',
+    );
+    expect(selectAuthMethod([], false)).to.equal(false);
+  });
+
+  it('preserves the public key authentication order', function () {
+    const selectAuthMethod = createSshAuthMethodSelector({
+      hasPassword: false,
+      hasPrivateKey: true,
+    });
+
+    expect(selectAuthMethod(null, null)).to.equal('none');
+    expect(selectAuthMethod(['publickey'], false)).to.equal('publickey');
+    expect(selectAuthMethod([], false)).to.equal(false);
+  });
+
+  it('preserves the combined password and public key order', function () {
+    const selectAuthMethod = createSshAuthMethodSelector({
+      hasPassword: true,
+      hasPrivateKey: true,
+    });
+
+    expect(selectAuthMethod(null, null)).to.equal('none');
+    expect(
+      selectAuthMethod(
+        ['password', 'publickey', 'keyboard-interactive'],
+        false,
+      ),
+    ).to.equal('password');
+    expect(
+      selectAuthMethod(['publickey', 'keyboard-interactive'], false),
+    ).to.equal('publickey');
+    expect(selectAuthMethod(['keyboard-interactive'], false)).to.equal(
+      'keyboard-interactive',
+    );
+    expect(selectAuthMethod([], false)).to.equal(false);
+  });
+
+  it('does not let methodsLeft change the current preference order', function () {
+    const selectAuthMethod = createSshAuthMethodSelector({
+      hasPassword: true,
+      hasPrivateKey: true,
+    });
+
+    expect(selectAuthMethod(null, null)).to.equal('none');
+    expect(selectAuthMethod(['publickey'], false)).to.equal('password');
+    expect(selectAuthMethod(['keyboard-interactive'], false)).to.equal(
+      'publickey',
+    );
+    expect(selectAuthMethod([], false)).to.equal('keyboard-interactive');
+  });
+
+  it('excludes keyboard-interactive after a partially successful password', function () {
+    const selectAuthMethod = createSshAuthMethodSelector({
+      hasPassword: true,
+      hasPrivateKey: false,
+    });
+
+    expect(selectAuthMethod(null, null)).to.equal('none');
+    expect(
+      selectAuthMethod(['password', 'keyboard-interactive'], false),
+    ).to.equal('password');
+    expect(selectAuthMethod(['keyboard-interactive'], true)).to.equal(false);
+  });
+
+  it('excludes keyboard-interactive after a partially successful public key', function () {
+    const selectAuthMethod = createSshAuthMethodSelector({
+      hasPassword: true,
+      hasPrivateKey: true,
+    });
+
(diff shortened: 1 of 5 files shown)


Anemy:
Hi @TemaSM, just letting you know we've seen this and it's on our radar. Thanks for opening the issue and the pr!

PR #854, opened 2026-08-13
chore(mongodb-schema): update test timeout on windows and ensure close COMPASS-10932

Anemy:
COMPASS-10932

Adding this package also shuffled which tests land in which test group (1-4). mongodb-schema is in group 3. I didn't deep dive all of the failures, on the two I did look at it was 3 timing out in mongodb-schema.

### DIFF
--- a/packages/mongodb-schema/test/integration/generate-and-validate.test.ts
+++ b/packages/mongodb-schema/test/integration/generate-and-validate.test.ts
@@ -54,8 +54,21 @@ describe('Documents -> Generate schema -> Validate Documents against the schema'
 });
 
 describe('With a MongoDB Cluster', function () {
+  if (process.platform === 'win32') {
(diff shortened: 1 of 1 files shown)


nbbeeken (approved):
~ 35min CI is tough but at least its green

Anemy:
@nbbeeken Looking at the logs, most of it is spent downloading mongodb on windows for mongodb-runner (and the packages that use mongodb-runner). Previously there were two packages that used it and they were split between 3 and  4. Now they're both on 4 so that means they're both downloading and it takes longer. mongodb-schema is now on 3 which is why that also takes a while (but not as long).

I'm going to merge this as it at least unblocks the CI passing. I'll keep the ticket open and open a pr that caches the windows mongodb download (if we can).

PR #855, opened 2026-08-13
fix(oidc-http-server-pages): bump compass-components version COMPASS-10980

esvm:
## Description

https://jira.mongodb.org/browse/COMPASS-10980

Bump the version of the `@mongodb-js/compass-components` lib as this was referencing a very old instance which was causing the callback from the login page to not render properly.

This issue was introduced when `leafgreen-ui/emotion` was bumped on [this change](https://github.com/mongodb-js/devtools-shared/commit/80755ed6d6e72d80c72d0ba16f2c2a0aec88ef7b). 

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

I have tested this by manually installing a local version of this lib into compass by running:
1. `cd packages/oidc-http-server-pages && npm pack`
2. In compass -> `npm install /path/to/devtools-shared/packages/oidc-http-server-pages/mongodb-js-oidc-http-server-pages-*.tgz`

## Checklist

### DIFF
--- a/packages/oidc-http-server-pages/package.json
+++ b/packages/oidc-http-server-pages/package.json
@@ -53,7 +53,7 @@
   },
   "devDependencies": {
     "@leafygreen-ui/emotion": "^5.2.0",
-    "@mongodb-js/compass-components": "^1.6.0",
+    "@mongodb-js/compass-components": "^1.68.0",
     "@mongodb-js/eslint-config-devtools": "^0.11.8",
     "@mongodb-js/mocha-config-devtools": "^1.1.3",
     "@mongodb-js/prettier-config-devtools": "^1.0.4",
@@ -68,6 +68,7 @@
     "eslint": "^7.25.0 || ^8.0.0",
     "gen-esm-wrapper": "^1.1.3",
     "mocha": "^8.4.0",
+    "mongodb-query-parser": "^4.7.16",
     "nyc": "^15.1.0",
     "prettier": "^3.8.1",
     "react": "^17.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2053,13 +2053,11 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.18.9",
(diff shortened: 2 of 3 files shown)


packages/oidc-http-server-pages/package.json
```diff
@@ -68,6 +68,7 @@
     "eslint": "^7.25.0 || ^8.0.0",
     "gen-esm-wrapper": "^1.1.3",
     "mocha": "^8.4.0",
+    "mongodb-query-parser": "^4.7.16",
```
esvm, line 71: alright without this the tests failed in the bootstrap due to this dependency being missed. Now that I added the depcheck says this is unused lol

esvm: image

nbbeeken (commented):
lockfile change seems large but I'm reading through it and it does all seem related to leafygreen and react so, must be component upgrade related

packages/oidc-http-server-pages/package.json
nbbeeken: you can update the depcheck config to ignore, but there's something odd there, looks like the script that needs it is in mongodb-ts-autocomplete, does that package depend on mongodb-query-parser? maybe it was accidentally getting it transiently. Adding it as an explicit dep of mongodb-ts-autocomplete seems like the way to solve it if that's true 🤔

esvm: only reference in the code I could find was in the `shell-bson-parser`. And funnily enough that package doesn't say it depends on `mongodb-query-parser`, but it also doesn't depend on the oidc-http or compass-components 🤔

esvm: and running bootstrap locally works just fine 🫠

nbbeeken (approved):

packages/oidc-http-server-pages/package.json
gribnoysup: Kind of a tangent, but looking at the reason compass-components depends on query-parser, we can probably remove this dependency, it's a bit out of pattern for other code there

PR #856, opened 2026-08-14
feat(mongodb-runner): add download dir as env, cache mongodb download on windows ci COMPASS-10932

Anemy:
Only uses them for a week so we avoid getting stale ones. Github automatically cleans up unused cache items. This does mean once a week CI will be that slower run.

### DIFF
--- a/packages/mongodb-runner/src/mongocluster.ts
+++ b/packages/mongodb-runner/src/mongocluster.ts
@@ -80,6 +80,8 @@ export interface RSMemberOptions {
 export interface CommonOptions {
   /**
    * Directory where server binaries will be downloaded and stored.
+   * Defaults to the `MONGODB_RUNNER_DOWNLOAD_DIR` environment variable if set,
+   * otherwise to `tmpDir`.
    */
   downloadDir?: string;
   /**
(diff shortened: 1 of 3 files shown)


packages/mongodb-runner/src/mongocluster.ts
```diff
@@ -380,7 +382,9 @@ export class MongoCluster extends EventEmitter<MongoClusterEvents> {
     cluster.defaultConnectionOptions = { ...options.internalClientOptions };
     if (!options.binDir) {
       options.binDir = await this.downloadMongoDb(
-        options.downloadDir ?? options.tmpDir,
+        options.downloadDir ??
+          process.env.MONGODB_RUNNER_DOWNLOAD_DIR ??
```
Anemy, line 386: Maybe we want to avoid using env here and instead pass the downloadDir everywhere? It will add a bit more code and could lead to folks not passing it though.

nbbeeken: I feel like env is fine but since this is in the CLI could we add it to the help menu along side the --downloadDir flag and call this PR a feat -> (feat to add env var control for the download location)

Anemy: Updated (and mentioned it in the readme).

nbbeeken (approved):

PR #858, opened 2026-08-18
fix(query-parser): update DBRef stringify to preserve type information COMPASS-10987

Anemy:
COMPASS-10987


packages/query-parser/src/stringify.ts
```diff
@@ -80,11 +80,13 @@ const BSON_TO_JS_STRING = {
...
+    // `toJSString` only returns undefined for values that stringify to
+    // nothing, which for an oid can only be `undefined` itself.
+    const oid = toJSString(v.oid, 0) ?? 'undefined';
     if (v.db) {
-      return `DBRef('${v.collection}', '${v.oid.toString()}', '${v.db}')`;
+      return `DBRef(${JSON.stringify(v.collection)}, ${oid}, ${JSON.stringify(v.db)})`;
     }
-
-    return `DBRef('${v.collection}', '${v.oid.toString()}')`;
+    return `DBRef(${JSON.stringify(v.collection)}, ${oid})`;
```
nbbeeken: How do we render the `fields` [argument](https://github.com/mongodb/js-bson/blob/8a9085df38a3c36648b69e4aea5323bf8056c55b/src/db_ref.ts#L41) if one exists on the `v`?

Anemy: We didn't. Nice catch !!! Added in https://github.com/mongodb-js/devtools-shared/pull/858/commits/10b78275e412ed643f68e81d0946e36657d0b0c2

nbbeeken (approved):

nbbeeken (approved):

PR #863, opened 2026-08-31
feat(connection-string-policy): add shared connection string policy checks VSCODE-821

ivandevp:
## Description

Adds new package to check connection string policy and be reusable across different clients (i.e. compass, vscode, mcp).

## Open Questions

- The `mongodb` peer dependency is pinned to a single major on purpose: the type-level checks that keep the lists exhaustive are only meaningful against one set of driver typings. Is this safe to assume?
- It was suggested that this could live in `devtools-connect` instead. I went with a separate package to avoid pulling its dependencies into a client that needs none of them, but happy to re-export it from there if preferred.

## Checklist

### DIFF
--- a/packages/connection-string-policy/README.md
+++ b/packages/connection-string-policy/README.md
@@ -1,0 +1,54 @@
+# @mongodb-js/connection-string-policy
+
+Checks a MongoDB connection string against a policy list of options, and reports the ones
+that fall outside it.
+
+Some connection string options change where credentials are sent, or whether the connection
+is verified. This package reports whether a connection string uses any of them, so that a
+tool can decide whether to accept it as-is, ask the user to confirm, or refuse it.
+
+It is not a connection string validator: it says nothing about whether a connection string
+is well-formed or whether connecting will succeed. It also does not decide what happens
+next — it returns data, never throws for policy reasons, and does no logging or prompting.
+
+## Usage
+
+```ts
+import { checkConnectionStringPolicy } from '@mongodb-js/connection-string-policy';
+
+const result = checkConnectionStringPolicy(connectionString);
+
+if (!result.withinPolicy) {
+  console.log(result.flaggedOptions);
+}
+```
+
+## What you get back
+
+| Field            | Meaning                                                                     |
+| ---------------- | --------------------------------------------------------------------------- |
+| `withinPolicy`   | `true` when nothing was flagged.                                            |
+| `flaggedOptions` | Names of the options that fall outside the policy list.                     |
+| `unparseable`    | `true` when the connection string could not be parsed; never within policy. |
+
+`flaggedOptions` contains option names only, never values, so it is safe to display or log
+without redacting it first. Two kinds of entry are not literal option names:
+
+- `'ssl/tls'` — the connection would not be TLS-protected (a non-SRV string to a non-local
+  host without `tls=true`, or an SRV string that sets `tls=false`).
+- `'authMechanismProperties.<NAME>'` — an individual auth mechanism property.
+
+## How the lists work
+
+Every connection string option known to the Node.js driver is listed as either allowed or
+disallowed, and two type-level functions make TypeScript fail to compile if an option
+appears in neither. That is deliberate: a new driver option being silently treated as
+acceptable is worse than the cost of maintaining the list, so a driver upgrade that adds an
+option is _supposed_ to break the build here.
+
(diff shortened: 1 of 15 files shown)


packages/connection-string-policy/src/connection-string-policy.ts
```diff
@@ -0,0 +1,285 @@
...
+  'raw',
+  'rejectUnauthorized',
+  'secureContext',
+  'secureProtocol',
+  'serializeFunctions',
+  'servername',
+  'session',
+  'tlsAllowInvalidCertificates',
+  'tlsAllowInvalidHostnames',
+  'tlsCAFile', // !
```
alenakhineika:
```suggestion
  'tlsCAFile',
```

packages/connection-string-policy/src/connection-string-policy.ts
```diff
@@ -0,0 +1,285 @@
...
+  'serverMonitoringMode',
+  'serverSelectionTimeoutMS',
+  'socketTimeoutMS',
+  'srvMaxHosts',
+  'srvServiceName',
+  'ssl', // Only if value is `true` or target host is local.
+  'timeoutMS',
+  'tls', // Only if value is `true` or target host is local.
+  'tlsCertificateKeyFile',
+  'tlsCertificateKeyFilePassword',
```
alenakhineika, line 74: Shouldn't we move `tlsCertificateKeyFile`/`tlsCRLFile` and the other local-file-reading options to the disallowed list? Or how do we make a decision on what goes in which bucket?

ivandevp: I know we tried moving it to the disallowed list but we ended up reverting it in https://github.com/mongodb-js/compass/commit/6fb15688868591f50572cf1b72f0c0c013a96a76 and I found [this thread](https://mongodb.slack.com/archives/G2L10JAV7/p1778062052728909?thread_ts=1778024558.206599&cid=G2L10JAV7) about it

packages/connection-string-policy/src/connection-string-policy.ts
```diff
@@ -0,0 +1,285 @@
...
+   * TLS-protected.
+   */
+  flaggedOptions: string[];
+}
+
+/**
+ * Check a connection string against the policy list, and report the options that fall
+ * outside it.
+ *
+ * This is *not* a general-purpose connection string validator: it makes no claim about
```
gribnoysup: Can we clarify what then it tells us?

packages/connection-string-policy/src/connection-string-policy.ts
gribnoysup: It's worth documenting somewhere around this code I think as this is not the first time this trips people up

packages/connection-string-policy/README.md
```diff
@@ -0,0 +1,54 @@
+# @mongodb-js/connection-string-policy
+
+Checks a MongoDB connection string against a policy list of options, and reports the ones
+that fall outside it.
+
+Some connection string options change where credentials are sent, or whether the connection
+is verified. This package reports whether a connection string uses any of them, so that a
```
addaleax: What does "verified" mean in this context? We should be clear about what types of conditions the validator checks: Whether connecting using this connection string is considered inherently unsafe, if it comes from an untrusted source. It doesn't tell you much about what happens after connecting, though

packages/connection-string-policy/src/connection-string-policy.ts
ivandevp: Fair — I'll rewrite it to lead with what it does tell you, whether the connection string only uses options that are safe to act on without asking the user first, and which options aren't.

packages/connection-string-policy/README.md
ivandevp: Will fix. "verified" was too vague here, I meant TLS certificate verification. I'll reframe it as you describe: the check is about whether connecting with this string is inherently unsafe when the string comes from an untrusted source, and say explicitly that it says nothing about what happens after connecting.

packages/connection-string-policy/src/connection-string-policy.ts
ivandevp: added some docs in code about this

addaleax (approved):

gribnoysup (approved):

alenakhineika (approved):

PR #871, opened 2026-09-02
chore(mongodb-runner): add @mongodb-js namespace to the package name

gribnoysup:
The only way for us to deal with ownership issues with this package is to rename it to have the mongodb-js namespace


Anemy (approved):

nbbeeken (changes requested):
I assume code-points was unintentional (as it usually is), just leaving req changes so we don't miss that


What happens to the `mongodb-runner` package now? is there a way we can make one last release to it where it is just a re-export of this one?

gribnoysup:
Yes, Anna asked me to adjust for that, although I think it's worth mentioning that as our main concern here is erroneous owners of the mongodb-runner package, re-exporting still leaves the potentially problematic dependency in the tree even if the version we know we published will be pointing to a new one. I'll push one small change related to that re-exporting shortly and will remove the accidentally committed file, thanks for spotting.

gribnoysup:
@nbbeeken [patch with the re-exported package](https://github.com/mongodb-js/devtools-shared/pull/872)

nbbeeken (approved):

nbbeeken:
Yea, my thinking is we migrate all the stuff we own still to the new namespaced package but at least the old one is a pointer in the interim while the community at large moves over. We should still deprecate `mongodb-runner` and discourage its use as a way of informing downstream

PR #626, opened 2026-03-18
feat(mongodb-runner): add `mongodb-runner ls --json`

addaleax:
Provide JSON-formatted output that contains serialized cluster data.

## Checklist
- [ ] I have signed the Contributor License Agreement (https://www.mongodb.com/legal/contributor-agreement)
- [ ] New tests and/or benchmarks are included
- [ ] Documentation is changed or added

PR #877, opened 2026-09-02
feat(mongodb-runner): add --json output to start and ls

autarch:
## Description

This adds a `--json` output option. When this is enabled, the only output the runner prints to stdout is a JSON object containing information about the started cluster. The JSON always contains an `id` and `connectionString` key.

This is handy for scripting with the runner, and makes it easy to save the cluster information in a structured way, which is particularly helpful if you want to save multiple clusters.

## Open Questions

None.

## Checklist

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/packages/mongodb-runner/README.md
+++ b/packages/mongodb-runner/README.md
@@ -50,6 +50,8 @@ Options:
       --docker       Docker image name to run server instances under    [string]
       --id           ID to save the cluster metadata under              [string]
       --all          for `stop`: stop all clusters                     [boolean]
+      --json         for `start` and `ls`: print machine-readable JSON
+                     on stdout instead of plain text                   [boolean]
       --debug        Enable debug output                               [boolean]
       --help         Show help                                         [boolean]
 ```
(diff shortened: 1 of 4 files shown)


gribnoysup:
I'd maybe suggest to take a look at https://github.com/mongodb-js/devtools-shared/pull/626. There's already some interfaces for serialization of these instances that should probably be used instead of building these from scratch right in the method

autarch:
> I'd maybe suggest to take a look at #626. There's already some interfaces for serialization of these instances that should probably be used instead of building these from scratch right in the method

@gribnoysup What's the state of #626? It would be good to build on top of that, but it's been open since March. I could fold those changes into this PR, which might be simplest.

gribnoysup:
@autarch 

> I could fold those changes into this PR, which might be simplest.

This works I think, this patch Anna opened is practically ready, was just waiting for feedback and got a bit lost over time, so would be good to use it as a starting point here

PR #881, opened 2026-09-07
chore: force bump patch version on all packages

gribnoysup:
Doing the force bump to test trusted publishing for all packages

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
     },
     "configs/eslint-config-devtools": {
       "name": "@mongodb-js/eslint-config-devtools",
-      "version": "0.11.8",
+      "version": "0.11.9",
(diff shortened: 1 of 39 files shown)


package-lock.json
```diff
@@ -34460,7 +34460,7 @@
       "requires": {
         "@mongodb-js/devtools-proxy-support": "^0.7.19",
         "@mongodb-js/mongodb-runner": "^6.12.0",
-        "@mongodb-js/oidc-http-server-pages": "1.2.14",
+        "@mongodb-js/oidc-http-server-pages": "^1.2.14",
```
gribnoysup, line 34463: Not sure why it was pinned, but this makes it too easy to mess up resolution in the monorepo, changing to caret range as all other packages

PR #882, opened 2026-09-07
feat(mocha-config-devtools)!: update mocha-config-devtools to support react 18, chai 4x, and remove enzyme - COMPASS-11105

dudaschar:
## Description

This PR is part of the React 18 upgrade in compass.

The initial scope was to remove enzyme and unlock react 18 support. 
When removing enzyme and updating compass with this PR version, an issue surfaced where the chai and sinon-chai version were not resulting in compatible versions anymore. The previous sinon-chai@4.0.1 requires chai@5 (which drops commonJS) and we where getting errors with chai@4.5. This is currently an issue in the main branch of this project, but compass is resolving to 1.0.4 (which uses sinon-chai@3.7) instead of 1.1.5 (latest).

Moving sinon-chai to 3.7 fixes this, since is the last version supporting chai 4+. Another problem surfaced when testing with compass, since thats no resolved chai 4.5 in the repo, and the duplicated copy was causing a few tests to break since because their features were not supported in 4.5 anymore. I changed chai to 4+, so the different packages will be able to resolve to their needed versions.

The other packages which also failed with the invalid chai@4.5 were updated to chai@3.7.

A follow up PR updating the @types/sinon-chai package will be opened, since they are not a breaking change but would cause a major bump in their packages if added here. 

Open to suggestions, if that's a better way to sort this.

https://jira.mongodb.org/browse/COMPASS-11105

## Checklist


packages/devtools-connect/package.json
```diff
@@ -83,7 +83,7 @@
     "os-dns-native": "^2.0.1",
     "prettier": "^3.8.1",
     "resolve-mongodb-srv": "^1.1.1",
-    "sinon-chai": "^4.0.1",
+    "sinon-chai": "^3.7.0",
```
gribnoysup: While we're at it, can we clean this up? As we do the sinon helpers registration in our mocha preset, we don't need to have dependency on it in these packages. Removing this dep (and chai.use calls) will make maintaining it a bit easier going forward

dudaschar: on it! I'll check what other packages we can do the same as well.

gribnoysup (approved):

PR #889, opened 2026-09-15
feat(monorepo-tools): allow single repo COMPASS-10984

esvm:
## Description

I'll be making use of the `bump-packages` script in https://github.com/mongodb-js/electron-wix-msi/.
Today, this script requires a workspace to exist, but that only makes sense for monorepos with multiple packages. 

I'm changing the script to allow a fallback to the root dir in case the `workspaces` is not defined in `package.json`.

**Testing**

- [x] Unit tests
- [x] Manual tests in the repo I want to use it

## Checklist


packages/monorepo-tools/src/utils/get-packages-in-topological-order.ts
```diff
@@ -19,10 +19,9 @@ export interface PackageInfo {
 export async function getPackagesInTopologicalOrder(
   monorepoRoot: string,
 ): Promise<PackageInfo[]> {
-  const patterns: string[] =
-    JSON.parse(
-      await fs.readFile(path.join(monorepoRoot, `package.json`), 'utf8'),
-    ).workspaces || [];
+  const patterns: string[] = JSON.parse(
+    await fs.readFile(path.join(monorepoRoot, `package.json`), 'utf8'),
+  ).workspaces || ['.']; // <---- default value that considers monorepo root as a "workspace" for further analysis
```
esvm:
```suggestion
  ).workspaces || ['.']; // default value that considers monorepo root as a "workspace" for further analysis
```

gribnoysup (approved):
Not sure if there's any documentation we might want to adjust also, code change lgtm 🙂

PR #892, opened 2026-09-21
feat(bson-parsing): restrict web worker scope COMPASS-11128

mabaasit:
## Description

[COMPASS-11128](https://jira.mongodb.org/browse/COMPASS-11128)

## Checklist

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


packages/shell-bson-parser/src/worker.ts
```diff
@@ -6,7 +6,102 @@ import {
 
 import type { WorkerRequest, WorkerResponse } from './worker-types.js';
 
-// Exported for test
+// Exported for tests
+export const ALLOWED_GLOBALS = new Set([
+  // Used by this file.
+  'self',
+  'onmessage',
+  'postMessage',
```
addaleax, line 14: Adding `const { self, onmessage, postMessage } = globalThis;` above this statement should remove this requirement, and we do want to avoid a situation where the "sandbox" could call `postMessage()`

packages/shell-bson-parser/src/worker.ts
```diff
@@ -6,7 +6,102 @@ import {
...
+  'undefined',
+  'NaN',
+  'Infinity',
+  'isNaN',
+  'isFinite',
+  'parseFloat',
+  'parseInt',
+  'encodeURIComponent',
+  'decodeURIComponent',
+  'Buffer',
```
addaleax, line 68: `Buffer` is a Node.js concept, do we need that?

packages/shell-bson-parser/src/worker.ts
```diff
@@ -6,7 +6,102 @@ import {
...
+}
+
+// Exported for tests
+export const DISALLOWED_PROTOTYPE_PROPS = [
+  '__proto__',
+  '__defineGetter__',
+  '__defineSetter__',
+  '__lookupGetter__',
+  '__lookupSetter__',
+  'constructor',
```
addaleax, line 90: `Object.prototype.constructor` is part of standard JS, why are we excluding that?

packages/shell-bson-parser/src/worker.ts
mabaasit: Yes, that's much better and will clean this up

